PR #146, opened 2020-03-09
no-restart

MarkDaoust:
Remove the recommendation to restart the notebook. This doesn't seem necessary, and cause confusion:

If I  restart the notebook after running `%tensorflow_version`, and then continue where I left off, then I'm back to tensorflow 1.15, which gives some unclear errror: `NotFoundError: /usr/lib/python3/dist-packages/tensorflow_quantum/core/ops/_tfq_simulate_ops.so: cannot open shared object file: No such file or directory`

### DIFF
--- a/docs/tutorials/barren_plateaus.ipynb
+++ b/docs/tutorials/barren_plateaus.ipynb
@@ -138,9 +138,7 @@
         "id": "FxkQA6oblNqI"
       },
       "source": [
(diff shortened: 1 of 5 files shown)


MichaelBroughton:
Ahhh good catch. Can you update this for all of the notebooks ?

MarkDaoust:
Done.

PR #147, opened 2020-03-09
`rm api_docs/`, these are on tensorflow.org now.

MarkDaoust:
These are published on tensorflow.org now.

This copy in github is not necessary for the site and is not kept up-to-date.

Let's delete it so nobody gets confused.


lamberta:
@MarkDaoust Just to confirm, best practice is to generate these and publish to a release branch for old versions, yes?

MarkDaoust:
> @MarkDaoust Just to confirm, best practice is to generate these and publish to a release branch for old versions, yes?

Generating a copy into a release branch or tag is something you can do, if you like being able to reference old versions. It just takes a little effort on your part to maintain.

We do this for tensorflow/docs, but have never really encouraged sub-projects to do it too. Some did it in the bast before we had the automation.
 
But having a copy on master that is not up to date is confusing.

lamberta:
Gotcha. Thanks!

MichaelBroughton:
Keeping things simple I think we’ll opt to just publish releases as branches and not bother to add in api_docs into those GitHub branches.

PR #156, opened 2020-03-11
Hello update

MichaelBroughton:
After getting a chance to see how people responded to the first tutorial, we thought it might be good if we added a little cell indicating how a user could inspect the results of their model.

I've added a new cell header called `Verifying outputs` which explicitly reviews how well the training of the model went and ties things back to the original problem we wanted to solve, which was correcting your qubit, given you didn't prepare it perfectly.


docs/tutorials/hello_many_worlds.ipynb
```diff
@@ -915,6 +917,72 @@
...
       ]
     },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "C2RfWismj66S",
+        "colab_type": "text"
+      },
+      "source": [
+        "### 2.6 Verifying outputs\n",
```
lamberta:
```suggestion
        "### 2.6 Verify outputs\n",
```

docs/tutorials/hello_many_worlds.ipynb
```diff
@@ -915,6 +917,72 @@
...
+        "Now that you have trained this model, you can use it to correct your system that has this calibration error. Using Cirq you can do this with:"
```
lamberta:
```suggestion
        "Now use the trained model to correct the qubit calibration errors. With Cirq:"
```

changed "system" to "qubit" to match the opening statement

docs/tutorials/hello_many_worlds.ipynb
```diff
@@ -915,6 +917,72 @@
...
+        "    state = cirq_simulator.simulate(\n",
+        "        full_circuit,\n",
+        "        {s:v for (s,v) in zip(control_params, params_to_prepare_output[index])}\n",
+        "    ).final_state\n",
+        "    expectation = z0.expectation_from_wavefunction(state, {qubit: 0}).real\n",
+        "    print(f'For a desired output (expectation) of {desired_values[index]} with'\n",
+        "          f' noisy preparation, the controller\\nnetwork found the following '\n",
+        "          f'values for theta: {params_to_prepare_output[index]}\\nWhich gives an'\n",
+        "          f' actual expectation of: {expectation}\\n')\n",
+        "\n",
```
lamberta: perhaps add another newline here to make the following call stand out. Or move the following to a separate cell

docs/tutorials/hello_many_worlds.ipynb
```diff
@@ -915,6 +917,72 @@
...
+      "outputs": []
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "wvW_ZDwmsws6",
+        "colab_type": "text"
+      },
+      "source": [
+        "If you aren't as concerned with the parameter values you could always just check the outputs from above using `tfq`:"
```
lamberta: > just check the outputs using `tfq`

Is this step obvious? If not, may want to provide a *small* amount of content

docs/tutorials/hello_many_worlds.ipynb
MichaelBroughton: Fixed.

docs/tutorials/hello_many_worlds.ipynb
MichaelBroughton: I like that a lot better, thanks!

docs/tutorials/hello_many_worlds.ipynb
MichaelBroughton: Added newline.

docs/tutorials/hello_many_worlds.ipynb
MichaelBroughton: Added a few more words here, let me know what you think ?

docs/tutorials/hello_many_worlds.ipynb
```diff
@@ -915,6 +917,73 @@
...
+      "outputs": []
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "wvW_ZDwmsws6",
+        "colab_type": "text"
+      },
+      "source": [
+        "The value of the loss function during training will give you a rough idea of how close to expect the **expectation values of the full circuit controlled by `controller` to be to the desired outputs** in the above cell. If you aren't as concerned with the parameter values you can always check the outputs from above using `tfq`:"
```
lamberta:
```suggestion
        "The value of the loss function—during training—provides a rough indication for how close to expect the expectation values of the full circuit, controlled by `controller`,  to the desired outputs in the above cell. If you aren't as concerned with the parameter values, you can always check the outputs using `tfq`:"
```
I tried to tighten up that first sentence but it's still pretty dense. Can you break it up into two?

MichaelBroughton: Broke it down. Thoughts ?

docs/tutorials/hello_many_worlds.ipynb
```diff
@@ -915,6 +917,73 @@
...
+      "outputs": []
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "wvW_ZDwmsws6",
+        "colab_type": "text"
+      },
+      "source": [
+        "The value of the loss function during training will give you a rough idea of how well the model is learning. **The lower the loss, the closer the expectation values in the above cell will be to `desired_values`**. If you aren't as concerned with the parameter values you can always check the outputs from above using `tfq`:"
```
lamberta:
```suggestion
        "The value of the loss function during training provides a rough idea of how well the model is learning. The lower the loss, the closer the expectation values in the above cell is to `desired_values`. If you aren't as concerned with the parameter values, you can always check the outputs from above using `tfq`:"
```

Yeah, much better. I made minor edit because I'm allergic to bold :)

PR #165, opened 2020-03-17
Some small util ops cleanup.

MichaelBroughton:
Cleaned up util ops and ps_util ops:
1. when dealing with tensor data apparently it's better to use `tensorflow::tstring` since it is more portable and such.
2. Added a few missing `OP_REQUIRES_OK`
3. Added shape outputs.


MichaelBroughton:
Good catch, changed them over now.

PR #167, opened 2020-03-18
Added symbol properties to high level layers.

MichaelBroughton:
Added a `symbols` property to the high level layers so that if users want to extract the values from the weight tensors they can use `symbols` to cross reference to existing symbols without having to do the sorting themselves.


MichaelBroughton:
Yes I think having a small function that zips together the result of symbols and self.get_weights makes a lot of sense! Do you want to take that on ?

zaqqwerty:
Sure, I will do that

PR #172, opened 2020-03-22
State space tests

### LINKED ISSUE #55: Test coverage of state spaces

zaqqwerty:
Resolves #55


tensorflow_quantum/core/qsim/state_space_test.cc
```diff
@@ -18,15 +18,380 @@ limitations under the License.
...
+  state_clone->CopyFrom(*state);
+
+  for (uint64_t i = 0; i < state->GetDimension(); i++) {
+    ASSERT_EQ(state->GetAmpl(i), state_clone->GetAmpl(i));
+  }
+
+  double real_inner_product = 0.0;
+  for (uint64_t i = 0; i < state->GetDimension(); i++) {
+    real_inner_product += 2 * i * i;
+  }
```
MichaelBroughton: Don't really need this for loop here.

tensorflow_quantum/core/qsim/state_space_test.cc
```diff
@@ -18,15 +18,380 @@ limitations under the License.
...
+    }
+  }
+}
+
+TEST(StateSpaceTest, ComputeExpectation) {
+  const uint64_t num_qubits(9);
+  const uint64_t q0(3);
+  const uint64_t q1(7);
+  auto state = std::unique_ptr<StateSpace>(GetStateSpace(num_qubits, 1));
+  auto scratch = GetStateSpace(num_qubits, 1);
```
MichaelBroughton: Why not put this in a smart pointer too ?

tensorflow_quantum/core/qsim/state_space_test.cc
```diff
@@ -18,15 +18,380 @@ limitations under the License.
...
+          tensorflow::Status(tensorflow::error::INVALID_ARGUMENT,
+                             "SSE simulator doesn't support small circuits."));
+      break;
+  }
+}
+
+TEST(StateSpaceTest, ApplyGate2) {
+  auto state = std::unique_ptr<StateSpace>(GetStateSpace(6, 1));
+  state->CreateState();
+  const float matrix_hi[] = {1.0 / std::sqrt(2),
```
MichaelBroughton: Can this formatting be more human readable ?

tensorflow_quantum/core/qsim/state_space_test.cc
```diff
@@ -18,15 +18,380 @@ limitations under the License.
...
 #include "tensorflow_quantum/core/qsim/mux.h"
+#include "tensorflow_quantum/core/src/circuit.h"
+
+#ifdef __AVX2__
+tfq::qsim::StateSpaceType STATE_SPACE_TYPE = tfq::qsim::StateSpaceType::AVX;
+#elif __SSE4_1__
+tfq::qsim::StateSpaceType STATE_SPACE_TYPE = tfq::qsim::StateSpaceType::SSE;
+#else
+tfq::qsim::StateSpaceType STATE_SPACE_TYPE = tfq::qsim::StateSpaceType::SLOW;
+#endif
```
MichaelBroughton: Since these are only used in the first test, what do you think about moving them inside of there and not making a global variable ?

tensorflow_quantum/core/qsim/state_space_test.cc
zaqqwerty: Replaced with a constant

tensorflow_quantum/core/qsim/state_space_test.cc
zaqqwerty: Moved inside of test

tensorflow_quantum/core/qsim/state_space_test.cc
zaqqwerty: Changed to smart pointer and used `get` to hand `ComputeExpectation` a raw pointer

tensorflow_quantum/core/qsim/state_space_test.cc
zaqqwerty: Reverted and added clang format guards

MichaelBroughton:
Looks good. As a final sanity check can you test locally where you run all tests without the sse flags and then finally include the avx flags ?

zaqqwerty:
Ran `scripts/test_all.sh` with and without all the sse flags and the tests pass.  For avx, what flags do you use?  I tried `--cxxopt="-mavx" --cxxopt="-mavx2"` but `state_space_avx.cc` doesn't compile

MichaelBroughton:
I think it also uses some FMA, try doing instead march=native.

zaqqwerty:
Everything passed when I added `-march=native`

Issue #144, opened 2020-03-09
Finish C++ Backend for SampledExpectation


MichaelBroughton:
This is a deceptive amount of work. Roughly we will need to do the following things:

1. Add ability to calculate sampled expectation in `qsim` and `src`.
2. Add a new C++ op that will have the same signature as the python.
3. Link that C++ op into `get_sampled_expectation_op`.

No need to modify any other tests or layers since this should just propagate up through the API.

MichaelBroughton:
Step 1 is done, before we move on to 2. #158 Should go in and we can start testing state_space a little more to make our lives easier for step 2.

zaqqwerty:
Resolved by #196

PR #181, opened 2020-03-31
[WIP] Upgrade to tf 2.2rc2 and bazel 2.0.0.

MichaelBroughton:
Still some questions up in the air:

1. I'm fairly confident we have bazel 2.0.0 working, but I am still a little worried about flags, does the C++ versioning or flags change for linux, windows or macos in going from 2.1.0 to 2.2rc2 ?

2. Can we confirm that colab will be rolling with tf2.2rc2 and it's safe to remove these install statements ? Should we just update them to `pip install tensorflow==2.2rc2` ?

3. Merging this will temporarily bork our nightly builds, so let's make sure we coordinate with fixing our nightly build pipelines before merging.

@yifeif Can you speak to 1. at all ?
@MarkDaoust Can you speak 2. at all ?
@jaeyoo Once we are ready to merge this I should be able to update the nightly pipelines but might need your help :)
@zaqqwerty Do you have anything you want to get in before this upgrade ?

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -76,16 +76,12 @@ http_archive(
 
 http_archive(
     name = "org_tensorflow",
-    sha256 = "e82f3b94d863e223881678406faa5071b895e1ff928ba18578d2adbbc6b42a4c",
-    strip_prefix = "tensorflow-2.1.0",
+    # sha256 = "e82f3b94d863e223881678406faa5071b895e1ff928ba18578d2adbbc6b42a4c",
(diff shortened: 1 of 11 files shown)


MarkDaoust:
Colab has  appears to have already rolled out`tensorflow==2.2.0-rc2`. So I think you're safe on that front.

zaqqwerty:
If this will be a new release then I think I should finish #144 first

yifeif:
Not that I know of for build flags for TF from 2.1.0 to 2.2rc2. Let us know if you are seeing issues though.

PR #187, opened 2020-04-02
Parity checker

zaqqwerty:
First piece of #144

### DIFF
--- a/tensorflow_quantum/core/qsim/util.cc
+++ b/tensorflow_quantum/core/qsim/util.cc
@@ -36,14 +36,5 @@ void* _aligned_malloc(size_t size) {
 
 void _aligned_free(void* ptr) { free(*(reinterpret_cast<void**>(ptr) - 1)); }
 
-inline int ComputeParity(const std::vector<unsigned int>& measured_bits, const uint64_t sample) {
-  uint64_t mask = 0;
-  for(unsigned int i = 0; i < measured_bits.size(); i++){
-    mask |= uint64_t(1) << uint64_t(measured_bits[i]);
-  }
-  int count = std::bitset<64>(sample & mask).count() & 1;
-  return count ? -1 : 1;
-}
-
 }  // namespace qsim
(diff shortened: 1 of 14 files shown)


tensorflow_quantum/core/qsim/util.cc
```diff
@@ -33,5 +35,21 @@ void* _aligned_malloc(size_t size) {
...
+    if (sample_copy % 2 == 1) {
+      if (parity_set.find(location) != parity_set.end()) {
+        parity *= -1;
+      }
+    }
+    location++;
+    sample_copy /= 2;
+  }
+  return parity;
+}
```
MichaelBroughton, line 52: Can this be simplified to something like this ?
```C++
inline int parity(const std::vector<int>& measured_bits, const uint64_t bitstring) {
	uint64_t mask = 0;
	for(int i =0;i<measured_bits.size();i++){
		mask |= uint64_t(1) << uint64_t(measured_bits[i]);
	}
	int count = std::bitset<64>(bitstring & mask).count() & 1;
	return count ? -1 : 1;
}
```
since the ith bit in `bitstring` will be set if qubit i measured 1 ?

zaqqwerty: Good point, this is a perfect candidate for bit operations, I've now put this code into the header

tensorflow_quantum/core/qsim/util.h
```diff
@@ -16,7 +16,10 @@ limitations under the License.
 #ifndef TFQ_CORE_QSIM_UTIL_H_
 #define TFQ_CORE_QSIM_UTIL_H_
 
+#include <bitset>
 #include <cstddef>
+#include <cstdint>
```
MichaelBroughton, line 21: Why is this needed ?

zaqqwerty: Seems it isn't, at some point I thought it was needed for uint64_t.  Removed.

Issue #185, opened 2020-04-02
Implement quantum execution context


MichaelBroughton:
For our slower (cirq_ops) ops where we have the potential of going through to engineSampler in Cirq we might have to drastically change certain execution parameters. For example things like:

1. Reducing from parallel execution to serial execution.
2. Increasing optimization for minimizing circuit depth.
3. Make better use of sweeps.
4. Grouping identical circuits in the batch into the same job.

Need to be defined at the global level. To implement this we should implement a context module in the `python` directory that allows you to set these parameters globally in a thread safe fashion. We can take pretty closely after this: https://github.com/tensorflow/tensorflow/blob/v2.1.0/tensorflow/python/eager/context.py

Implementing this should roughly involve:
1. Implement `quantum_context.py` and tests
2. Implement `engine_util.py` and tests (#186 )
3. Connect `engine_util.py` to `cirq_ops.py` via `quantum_context.py`.
4. Allow users to enable or disable engine compatability mode via `tf.enable_engine_mode` in `util.py`

zaqqwerty:
See the discussion [here](https://github.com/tensorflow/quantum/pull/316#issuecomment-667722886)

zaqqwerty:
For further clarification of my questions under #188 , it seems that the goals listed can all be achieved without adding a global state:
1) I think the choice between parallel and serial execution should be with the cirq object doing the expectation values or sampling, and TFQ should be able to query these objects in a way that is agnostic the choice
2) Minimizing circuit depth is a circuit transformation, so it seems that if this is to be a capability of TFQ then it should be an op ingesting and emitting circuits, rather than something buried in the call to engine.  This seems like something that should be explicit and user-facing because minimizing circuit depth is not a transformation specific to engine.
3) Am I right to assume that "Make better use of sweeps" means something like: identifying when circuits in a batch are identical and calling `run_sweep` with a batch of parameter resolvers instead of calling `run` for each circuit, in order to eliminate the superfluous latency incurred when asking engine to run circuits one-by-one?  If so then this also seems like it should be circuit execution backend specific, and should be something configurable on initialization (that is, these functions would be handled by some TFQ op or Cirq class, rather than being activated by a global state).
4) If identical means the parameters are also the same, then it seems this would benefit any backend, so if we have a good way to do this, we should always do it, rather than only for engine.

MichaelBroughton:
[...]

PR #188, opened 2020-04-03
Added quantum_context.

MichaelBroughton:
First step in #185 . From here we will need to implement `engine_util.py` and then hook that into `cirq_ops.py` just like `batch_util.py` but for engine queries.


zaqqwerty:
I want to get a better sense of what the eventual user experience might look like.  Based on the TF docs, it seems that the standard for setting models to train on heterogeneous hardware is the [`tf.distribute.Strategy`](https://www.tensorflow.org/guide/distributed_training) object, which seems to take care of sending model pieces to appropriate places.  I have also seen the ability to set which devices a function runs on too, [for example](https://www.tensorflow.org/guide/gpu#manual_device_placement):
```
# Place tensors on the CPU
with tf.device('/CPU:0'):
  a = tf.constant([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]])
  b = tf.constant([[1.0, 2.0], [3.0, 4.0], [5.0, 6.0]])
```
Based on this I would imagine running a quantum op on QPU would look like:
```
with tf.device('/QPU:0'):
  expectations = tfq.get_expectation_op()(
      my_circuit_tensor, ['alpha'], my_values, my_paulis)
```
Is this the direction things will go?

MichaelBroughton:
In it's current implementation no. `tf.devices` usually dictate an explicit choice of underlying devices in C++ OpKernels. For example if I wrote an op that was defined on both CPU and GPU I could then explicitly choose to use one or the other.

In theory we could add a new device called qpu and allow the user to register C++ ops that would "run" on the QPU by using engine api calls or RPC calls. However that would limit us in some serious ways: We wouldn't have access to any of the nice machinery that Cirq provides for making efficient engine queries for things like expectation calculation since we would have to write this op in C++. This could be worthwhile once engine API's and best practices become a little more cemented in place.

Right now hitting engine can go something like this:

```python
import cirq

tfq.enable_engine_mode()
engine_sampler = <A complicated setup for engine that does depend on user input>
tfq.layers.Sampple(backend=engine_sampler)(cirq.Circuit(...))

# Engine enabled mode only applies to python backends so something like:
tfq.layers.Sample()(cirq.Circuit(...))
# is not affected

# this would be affected
tfq.layers.Sample(backend=cirq.Simulator())(cirq.Circuit(...))
```
In theory we could switch to using the devices, but to remove ambiguity we would need to do things like:
1. Write more simulators in C++ to support the full range of cirq simulators that could be plugged in for `backend`. This would mean: density matrix, clifford 
2. Remove cirq_ops
3. write a C++ op to efficiently hit engine for sampling and expectation.

Since these are big undertakings it might make sense to first look at if we should take the time to add fast density matrix simulation, cliffords etc.

MichaelBroughton:
I would also be in favor of the C++ op approach, but we are pretty limited badnwidth wise and moving to that would be an enormous amount of work for not a lot of gain. There wouldn't be major performance boosts and there would just be an API change to using tf.devices instead of this `tf.enable_engine_mode`. We would also lose the ability to support new cirq.Sampler interfaces and `cirq.SimulatesFinalState` interfaces were they to be added.

I'd say overall C++ might be better, but the isn't much performance or user benefit to doing it now, it's a lot of work and there are other lower hanging problems in the meantime.

MichaelBroughton:
I'm going to chat with some other TF engineers offline to see what their take is on things. Will report back.

MichaelBroughton:
Going to update and put up for review.

zaqqwerty:
You mentioned you might chat with some TF engineers about things here, did they have any feedback?

MichaelBroughton:
They confirmed that devices are not what we want here and that these global configurations that can be configured right after import are the way to go for swapping over op behavior etc.

zaqqwerty:
I realize I don't understand why the global traits enabled by this quantum context can't be carried around by the cirq engine object itself instead.  For example, what will `set_low_latency_op_mode(mode)` do that can't be a configuration option of the engine before passing it into a TFQ layer?

I am imagining the following scenario: multiple `tfq.layers.Expectation` objects are run in separate branches of a Keras functional model; they may be running on different machines, so they might all call out to engine simultaneously with different requests; this might overload the engine.  Is the quantum context supposed to address this?  If so, I would disagree that TFQ is supposed to handle this, it seems like Engine should be written more carefully to accept jobs in a buffer so as to not be overloaded.

zaqqwerty:
This dovetails with our discussion under #316

MichaelBroughton:
> I imagine it might be to address the following scenario: multiple tfq.layers.Expectation objects are run in separate branches of a Keras functional model; they may be running on different machines, so they might all call out to engine simultaneously with different requests; this might overload the engine. Is the quantum context supposed to address this? If so, I would disagree that TFQ is supposed to handle this, it seems like Engine should be written more carefully to accept jobs in some kind of buffer so as to not be overloaded.

It also adresses different situations where global modifications to behaviors are required. An example similar to this one would be: We have lots of `tfq.layers.Expectation` layers in a compute graph and want to pump 35 qubit circuits through them. With TensorFlow's existing parallelization scheme it would try to run ALL the layers in parallel with sharded compute resources. If this happens to take place on one machine that means our memory would jump to `num_expectation_layers * size of state vector for 35 qubit circuit` which would effectively destroy any possibility of running. The second intent of `low_latency_op_mode` is to prevent this from happening by having layers encapsulate high memory computations with `tf.CriticalSection().execute` in order to manage memory down to just `size of state vector for 35 qubits` regardless of how many layers we might have.

Re: making engine faster to support TF's aggressive parallelization tendencies. Right now we pump in so many queries that we basically deadlock either ourselves or engine ( or both ) with no hope of recovering. That's to graph execution parallelization, even if we wrote a totally serial op for querying engine and had too many of them in a given compute graph it could still deadlock. We need the additional capability of locking down TF's parallelization AND locking down in-op parallelism. `low_latency_op_mode` addresses the former (which also tackles the above case) and  `engine_mode` addresses the latter.

zaqqwerty:
Re "The second intent of low_latency_op_mode is to prevent this from happening by having layers encapsulate high memory computations with `tf.CriticalSection().execute` in order to manage memory down to just `size of state vector for 35 qubits` regardless of how many layers we might have.", why can't we just always do this (since our ops can check for qubit number)?  More broadly, if this is something that is only required in some scenarios, it seems brittle to leave this decision up to the user adding a global state, rather than having an automatic solution.

Too many calls to engine causing it to deadlock seems to me like an issue that is up to the engine to solve, could you give a bit more context around why this is insoluble there?  Even if this is an issue with calls to engine, why can't the python `Engine` object initialization set up whatever global state it needs itself to prevent this from happening?

zaqqwerty:
Just saw your comment on #185, sorry this got a bit fragmented across threads

PR #189, opened 2020-04-03
Sampled expectation calculator

zaqqwerty:
Next step towards #144.  Adds `ComputeSampledExpectation` to statespaces.

### DIFF
--- a/tensorflow_quantum/core/qsim/state_space.cc
+++ b/tensorflow_quantum/core/qsim/state_space.cc
@@ -193,8 +193,7 @@ tensorflow::Status StateSpace::ComputeSampledExpectation(
 
     // Compute the expectation value over the samples
     int parity_total(0);
-    uint64_t parity_mask;
-    parity_mask = ComputeBitmask(parity_bits);
+    uint64_t parity_mask = ComputeBitmask(parity_bits);
     for (const uint64_t& sample : state_samples) {
       parity_total += ComputeParity(parity_mask, sample);
     }
--- a/tensorflow_quantum/core/qsim/state_space.cc
+++ b/tensorflow_quantum/core/qsim/state_space.cc
@@ -191,7 +191,8 @@ tensorflow::Status StateSpace::ComputeSampledExpectation(
             tensorflow::error::INVALID_ARGUMENT,
             "Could not parse Pauli term qubit id: " + pair.ShortDebugString());
       }
-      parity_bits.push_back(location);
+      // Parity functions use little-endian indexing
+      parity_bits.push_back(num_qubits_ - location - 1);
     }
 
     // Compute the expectation value over the samples
--- a/tensorflow_quantum/core/qsim/state_space.cc
+++ b/tensorflow_quantum/core/qsim/state_space.cc
@@ -193,8 +193,10 @@ tensorflow::Status StateSpace::ComputeSampledExpectation(
 
     // Compute the expectation value over the samples
     int parity_total(0);
+    uint64_t parity_mask;
+    parity_mask = ComputeBitmask(parity_bits);
     for (const uint64_t& sample : state_samples) {
-      parity_total += ComputeParity(parity_bits, sample);
+      parity_total += ComputeParity(parity_mask, sample);
     }
     *expectation_value += static_cast<float>(parity_total) *
                           term.coefficient_real() / static_cast<float>(m);
--- a/tensorflow_quantum/core/qsim/state_space.cc
+++ b/tensorflow_quantum/core/qsim/state_space.cc
@@ -23,6 +23,7 @@ limitations under the License.
 #include "tensorflow/core/lib/random/simple_philox.h"
 #include "tensorflow_quantum/core/proto/pauli_sum.pb.h"
 #include "tensorflow_quantum/core/qsim/fuser_basic.h"
+#include "tensorflow_quantum/core/qsim/util.h"
 #include "tensorflow_quantum/core/src/circuit.h"
 #include "tensorflow_quantum/core/src/circuit_parser.h"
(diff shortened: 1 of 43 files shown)


tensorflow_quantum/core/qsim/state_space.cc
```diff
@@ -149,6 +150,58 @@ void StateSpace::SampleState(const int m, std::vector<uint64_t>* samples) {
...
+    scratch->SampleState(m, &state_samples);
+
+    // Find qubits on which to measure parity
+    std::vector<unsigned int> parity_bits;
+    for (const tfq::proto::PauliQubitPair& pair : term.paulis()) {
+      unsigned int location;
+      if (!absl::SimpleAtoi(pair.qubit_id(), &location)) {
+        return tensorflow::Status(
+            tensorflow::error::INVALID_ARGUMENT,
+            "Could not parse Pauli term qubit id: " + pair.ShortDebugString());
```
MichaelBroughton, line 189: Not entirely sure we need this since any PauliSum objects coming in most likely came from `ResolveQubitIds` which already puts things in the right format, but this isn't a huge deal I guess.

tensorflow_quantum/core/qsim/state_space.cc
```diff
@@ -149,6 +150,58 @@ void StateSpace::SampleState(const int m, std::vector<uint64_t>* samples) {
...
+      }
+      parity_bits.push_back(location);
+    }
+
+    // Compute the expectation value over the samples
+    int parity_total(0);
+    for (const uint64_t& sample : state_samples) {
+      parity_total += ComputeParity(parity_bits, sample);
```
MichaelBroughton, line 197: This looks like it is doing unescessary work. ComputeParity will always recompute the mask from `parity_bits` despite it being the same over and over again. Maybe we should merge the ComputeParity function into here to avoid that ? 

Runtime here is `n_terms * (m*time to compute mask + m*bitwise parity cost)`
Could be reduced to `n_terms * (m * bitwise parity cost + time to compute mask)`

tensorflow_quantum/core/qsim/state_space.cc
```diff
@@ -149,6 +150,58 @@ void StateSpace::SampleState(const int m, std::vector<uint64_t>* samples) {
...
+    }
+    *expectation_value += static_cast<double>(parity_total) *
+                          term.coefficient_real() / static_cast<double>(m);
```
MichaelBroughton, line 200: These should be floats

tensorflow_quantum/core/qsim/state_space_test.cc
```diff
@@ -466,6 +466,230 @@ TEST(StateSpaceTest, SampleStateComplexDist) {
   ASSERT_EQ(samples.size(), m);
 }
 
+TEST(StateSpaceTest, ComputeSampledExpectation) {
+  const uint64_t num_qubits(2);
+  const uint64_t q0(0);
+  const uint64_t q1(1);
+  const int m(10000000);
```
MichaelBroughton, line 473: Could this be safely reduced to 100k instead of 10M ?

tensorflow_quantum/core/qsim/state_space.cc
zaqqwerty: I find without a check we get the warning `ABSL_MUST_USE_RESULT bool SimpleAtoi(absl::string_view s, int_type* out)`

tensorflow_quantum/core/qsim/state_space.cc
zaqqwerty: Changed to float

tensorflow_quantum/core/qsim/state_space.cc
MichaelBroughton: Fair enough, can't hurt to do an extra safety check here and there.

tensorflow_quantum/core/qsim/state_space_test.cc
zaqqwerty: Lowered precision and num samples

tensorflow_quantum/core/qsim/state_space.cc
zaqqwerty: Added a new inline function to compute the bitmask only once

tensorflow_quantum/core/qsim/state_space_test.cc
```diff
@@ -466,6 +466,230 @@ TEST(StateSpaceTest, SampleStateComplexDist) {
   ASSERT_EQ(samples.size(), m);
 }
 
+TEST(StateSpaceTest, ComputeSampledExpectation) {
```
MichaelBroughton, line 469: Can we add cases for m=0 and m=1 ?

tensorflow_quantum/core/qsim/state_space.cc
```diff
@@ -149,6 +150,60 @@ void StateSpace::SampleState(const int m, std::vector<uint64_t>* samples) {
...
+            tensorflow::error::INVALID_ARGUMENT,
+            "Could not parse Pauli term qubit id: " + pair.ShortDebugString());
+      }
+      parity_bits.push_back(location);
+    }
+
+    // Compute the expectation value over the samples
+    int parity_total(0);
+    uint64_t parity_mask;
+    parity_mask = ComputeBitmask(parity_bits);
```
MichaelBroughton, line 197: Can do this in one line.

zaqqwerty: Changed to one-liner

tensorflow_quantum/core/qsim/state_space_test.cc
zaqqwerty: Added test for the cases m=0 and m=1

PR #192, opened 2020-04-06
Localize endian conversions

zaqqwerty:
Enforced big-endian consistency everywhere except in qsim statespace implementations.  Made matrix mul big-endian in `matrix` files and move them to qsim module since they are only used there (after gate fusion).


zaqqwerty:
Decided it made sense to put it with Gates since it is only used in `gates_def`.  Happy to put it back if this doesn't seem warranted

MichaelBroughton:
I think it would make more sense to put it back where it was. It's a more broadly applicable matrix function that might emerge in other places like if we ever implement a batched version of `cirq.unitary` or something like that.

zaqqwerty:
I put `Matrix4Permute` back

PR #194, opened 2020-04-08
Cleaned up SSE expectation.

MichaelBroughton:
Cut down the number of load instructions by 2x and also simplified some of the logic inside. Having those extra load instructions was also causing us to go OOB by 8 bytes on the last iteration of the loop.

### DIFF
--- a/tensorflow_quantum/core/qsim/state_space_sse.cc
+++ b/tensorflow_quantum/core/qsim/state_space_sse.cc
@@ -114,10 +114,9 @@ void StateSpaceSSE::SetStateZero() {
 
 float StateSpaceSSE::GetRealInnerProduct(const StateSpace& other) const {
   uint64_t size2 = GetDimension() / 4;
-  __m128d expv_0 = _mm_setzero_pd();
-  __m128d expv_1 = _mm_setzero_pd();
-  __m128d temp = _mm_setzero_pd();
+  __m128d expv = _mm_setzero_pd();
   __m128d rs_0, rs_1, is_0, is_1;
+  __m128 load_reg;
 
   auto statea = GetRawState();
   auto stateb = other.GetRawState();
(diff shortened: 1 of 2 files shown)


tensorflow_quantum/core/qsim/state_space_sse.cc
```diff
@@ -126,37 +125,40 @@ float StateSpaceSSE::GetRealInnerProduct(const StateSpace& other) const {
   // Currently not a thread safe implementation of inner product!
   for (uint64_t i = 0; i < size2; ++i) {
     // rs = _mm256_cvtps_pd(_mm_load_ps(statea + 8 * i));
-    rs_0 = _mm_cvtps_pd(_mm_load_ps(statea + 8 * i));
-    rs_1 = _mm_cvtps_pd(_mm_load_ps(statea + 8 * i + 2));
+    load_reg = _mm_load_ps(statea + 8 * i);
```
zaqqwerty, line 128: Trying to understand how things are organized here, I'll describe what I think is happening for you to confirm.  My understanding is that `statea` is storing blocks of 8 floats at each `i` index, where the last 4 are always zero.  Then `_mm_load_ps(statea + 8*i)` loads the first 4 floats of the block; call them `a`, `b`, `c`, and `d`.  Further, `b` and `d` are always zero.  Then, `rs_0` contains all 4 floats, packed in pairs to doubles, `ab` and `cd`.  Rather differently, `rs_1` contains 2 copies of the the second two floats, as `cd` and `cd`.  So at this point `rs_0` contains the real and imaginary parts of the state amplitude at index `i`, while `rs_1` contains two copies of the the imaginary part.  Is that right?

zaqqwerty: Thanks, that makes sense to me now

PR #196, opened 2020-04-09
Sampled expectation backend

zaqqwerty:


zaqqwerty:
One thing I am concerned about, and would like your opinion @MichaelBroughton on how to address, is that I had to change line 33 of `differentiators/linear_combination_test.py` from `2e3` samples to `3e6` samples to get the last test in that file to pass.  Could this be from an insufficient RNG?

MichaelBroughton:
@jaeyoo might know better than me why that is happening.

jaeyoo:
@zaqqwerty @MichaelBroughton Oops, sorry I just saw this PR. In my experience of implementing the tests, the extraordinary large number of samples was due to 1) not insufficient RNG + limit of float precision, or 2) not exact implementation itself. For 2), I had experience that somewhere terms were sampled duplicated, and terms that should have been sampled were not sampled etc. This kind of mistake in implementation was hard to be found because averaging them with the larger number of samples eliminated them, and it eventually seemed to work.

It could be one of 1) and 2), please let me review this PR by this weekend. Thanks a lot :)

PR #197, opened 2020-04-09
Corrected metalearning example

### LINKED ISSUE #191: Minor bug in Meta-Learning QAOA notebook
There's a line in the call function of the QRNN class that I think is incorrect. `prev_input` should be assigned `inputs[4]` instead of `inputs[3]`. Code shown below with stars around the line in question.
```
class QRNN(tf.keras.layers.Layer):
  def __init__(self, symbol_names):
    super(QRNN, self).__init__()
    self.shared = tf.keras.layers.Dense(25, name="shared")
    self.state = tf.keras.layers.Dense(25, name="state")
    self.params = tf.keras.layers.Dense(2, name="params")
    self.expectation = tfq.layers.Expectation()
    self.symbol_names = symbol_names

  def call(self, inputs):
    circuits = inputs[0]
    ops = inputs[1]
    state = inputs[2]
    params = inputs[3]
    **prev_output = inputs[3]**
    joined = tf.keras.layers.concatenate([state, params, prev_output])
    shared = self.shared(joined)
    s_inp = self.state(shared)
    p_inp = self.params(shared)
    exp_out = self.expectation(circuits,
                               symbol_names=self.symbol_names,
                               symbol_values=p_inp,
                               operators=ops)
    return [circuits, ops, s_inp, p_inp, exp_out]
```

zaqqwerty:
There were some surprising typos, which I have now corrected.  Resolves #191.

### DIFF
--- a/metalearning_qaoa/metalearning_qaoa.ipynb
+++ b/metalearning_qaoa/metalearning_qaoa.ipynb
@@ -72,7 +72,7 @@
         "\n",
         "Created : 2020-Feb-06\n",
         "\n",
-        "Last updated : 2020-Mar-02"
+        "Last updated : 2020-Apr-09"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


zaqqwerty:
Yeah I ran it all the way through

PR #212, opened 2020-04-20
x_train_bin may also be filtered

tiancheng2000:
Both `qnn` and `fair_nn` are based on binary encoded `x_train_bin`, however this dataset is full of contradictory examples (7789 out of 11520). Though already mentioned in the doc, "binarizing will cause more collisions", it might be unclear to a part of readers that the ratio of collision exceeded 67.6%! and this ratio will affect classificaion training.


MichaelBroughton:
Gentle reminder @MarkDaoust

PR #220, opened 2020-04-23
Input check refactor

zaqqwerty:
Factors out code common to the 4 circuit execution layers


tensorflow_quantum/python/layers/circuit_executors/input_checks.py
```diff
@@ -0,0 +1,157 @@
...
+                            " must be a string or sympy.Symbol.")
+        symbol_names = [str(s) for s in symbol_names]
+        if not len(symbol_names) == len(list(set(symbol_names))):
+            raise ValueError("All elements of symbol_names must be unique.")
+        symbol_names = tf.convert_to_tensor(symbol_names,
+                                            dtype=tf.dtypes.string)
+    elif tf.is_tensor(symbol_names):
+        if not symbol_names.dtype == tf.dtypes.string:
+            raise TypeError("symbol_names tensor must have dtype string.")
+        if not symbol_names.shape[0] == len(list(set(symbol_names.numpy()))):
```
MichaelBroughton, line 71: These checks rely on having access to `.numpy()` and `.shape[0]`. In the past these have failed when wrapped in `@tf.function` despite passing bazel tests. These tests are also doing data type checking which is done at the op level so it might be best to just revert this block to how it looked before.

tensorflow_quantum/python/layers/circuit_executors/input_checks.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # Ingest and promote symbol_values.
+    if isinstance(symbol_values, (list, tuple, np.ndarray)):
+        symbol_values = tf.convert_to_tensor(symbol_values,
+                                             dtype=tf.dtypes.float32)
+    elif tf.is_tensor(symbol_values):
+        if not symbol_values.dtype == tf.dtypes.float32:
+            raise TypeError("symbol_values tensor must have dtype float32.")
+    else:
+        raise TypeError("symbol_values must be list-like.")
```
MichaelBroughton, line 84: Same for here.

tensorflow_quantum/python/layers/circuit_executors/input_checks.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # Ingest and promote circuit.
+    symbol_batch_dim = tf.gather(tf.shape(symbol_values), 0)
+    if isinstance(inputs, cirq.Circuit):
+        # process single circuit.
+        inputs = tf.tile(util.convert_to_tensor([inputs]), [symbol_batch_dim])
+    # TODO(zaqqwerty): util.convert_to_tensor does not accept ndarrays
```
MichaelBroughton, line 91: This is actually intended behavior. ndarrays containg cirq circuits will tend to partially unpack the circuit in weird ways since circuits are iterable.

tensorflow_quantum/python/layers/circuit_executors/input_checks.py
```diff
@@ -0,0 +1,157 @@
...
+    elif isinstance(inputs, (list, tuple)):
+        # process list of inputs.
+        inputs = util.convert_to_tensor(inputs)
+    elif tf.is_tensor(inputs):
+        if not inputs.dtype == tf.dtypes.string:
+            raise TypeError("inputs tensor must contain serialized circuits.")
```
MichaelBroughton, line 97: Should let ops take care of this. Also very brittle as a tensor filled with ['junk'] will slip through here.

tensorflow_quantum/python/layers/circuit_executors/input_checks.py
```diff
@@ -0,0 +1,157 @@
...
+        operators = util.convert_to_tensor([[operators]])
+        op_needs_tile = True
+    elif isinstance(operators, (list, tuple)):
+        if not isinstance(operators[0], (list, tuple)):
+            operators = [operators]
+            op_needs_tile = True
+        operators = util.convert_to_tensor(operators)
+    elif tf.is_tensor(operators):
+        if not operators.dtype == tf.dtypes.string:
+            raise TypeError("operators tensor must contain serialized paulis.")
```
MichaelBroughton, line 146: It's also easy for `['junk']` to slip in here.

zaqqwerty:
I copied the logic from the `state` layer into `expand_circuits`, and removed `np.ndarray` from the list of allowed types for `inputs` and `symbol_names`; then copied the `Ingest and promote operators` logic from `expectation` layer into `expand_operators`, and removed `np.ndarray` from the list of allowed types for `operators`; then removed the no longer applicable test cases in `input_checks_test`.

PR #222, opened 2020-04-29
Uoft data

therooler:
Hi,

This is the code I wrote to load our data into TFQ. It consists of the following parts:

1. A `SpinSystem` class that handles a single instance of a N spin system for a single order parameter `g`.
2. A `SpinSystemDataset` class that contains `m` `SpinSystem`'s, where `m` is the number of order parameters for which we solved this particular system.
3. A `TFIChain` class that instantiates the `SpinSystemDataset`'s for the transverse field Ising-model for N=4,8,12,16 on a closed chain. 
4. The tests just check the interface and verify that the `cirq.Simulator()` state has >99.9% overlap with the ground state.

I added the classes for the other data we have as well, but commented them out for now.

Right now this code works if the data is located at the hardcoded path `/tmp/data/`, which has the following structure:
```
/TFI_closed_4
/TFI_closed_8
    ->/0.20
    ->/0.22
    ->/0.24
         ->/energies.npy
         ->/energy.npy
         ->/final_state.npy
         ->/groundstate.npy
         ->/H.npz
         ->/params.npy
         ->/stats.txt
            ...
       ...
```
All of this only works for systems with a single order parameter, which should be fine for now. Maybe at some point we want to add a 2D, or 3D grid of parameters for the XY-model, and then things will have to be adapted slightly.

@MichaelBroughton mentioned that it would be best if the data could be generated by the user. Although this is possible for the Hamiltonians, ground states and ground state energies (I have some code for this as well), generating the circuit parameters would require the user to run the full VQE optimization for each order parameter. So I think downloading the parameters is probably necessary. Luckily, these aren't VGG-size parameter sets. For instance, all 80 parameter sets for the 8 qubit TFI chain totals 15 kb of data.

Let me know what you think.

Regards,
Roeland


MichaelBroughton:
@lamberta What do you think is the best way to handle data downloading ?

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,240 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Quantum datasets for several quantum many-body spin systems in the form of
+parameterized circuits"""
+from typing import List
```
MichaelBroughton, line 17: I do appreciate the use of typing, but unlike Cirq we don't type check our code and to keep things consistent it might be better if we removed it for now.

lamberta:
@MarkDaoust WDYT?
Typically for datasets in core docs we try and get them in [TFDS](https://www.tensorflow.org/datasets/catalog/overview), and other stuff we host in our GCP bucket. Will [tfq.datasets](https://www.tensorflow.org/quantum/api_docs/python/tfq/datasets) be a TFDS-like hub specialized for quantum datasets? Can you uses TFDS and add a "quantum" category to the catalog?

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,240 @@
...
+                SpinSystem(g, res_e, fidelity, gs_energy, parameters,
+                           path + directory))
+        return sorted(systems)
+
+    def _download_dataset(self, path: str):
+        # TODO: Implement this.
+        pass
+
+
+class TFIChain(SpinSystemDataset):
```
MichaelBroughton, line 134: Could we turn this into a function that returns a tuple of lists/arrays to the effect of ?
```python
def tfi_chain(...):
    .
    .
    .
    return circuits(with parameters resolved in), g, gs_energy, res_e
```
Where `circuits[i]` has corresponding information `g[i]`, `gs_energy[i]` or `res_e[i]`. This feels more natural for a user reading the documentation who could then do something like:
```python
circuits, g, gse, _ = tfq.datasets.tfi_chain(...)
data_tensor = tfq.convert_to_tensor(circuits)
label_tensor = tf.convert_to_tensor(gse)
```

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def hamiltonian(self) -> sps.csr_matrix:
+        """Load the hamiltonian into a scipy sparse CSR matrix."""
+        return sps.load_npz(self.path + '/H.npz')
+
+    def ground_state(self) -> np.ndarray:
+        """Load the ground state wave function into a complex numpy array."""
+        return np.load(self.path + '/groundstate.npy')
+
+    def final_state(self):
```
MichaelBroughton, line 63: Do we need this if a user can do cirq.final_wavefunction(the_circuit(parameters)) to get this ?

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,240 @@
...
+        self.res_e = res_e
+        self.fidelity = fidelity
+        self.gs_energy = gs_energy
+        self.parameters = parameters
+        self.path = path
+
+    def __lt__(self, other):
+        return self.g < other.g
+
+    def hamiltonian(self) -> sps.csr_matrix:
```
MichaelBroughton, line 55: Would we be able to return a `cirq.PauliSum` object here instead ?

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,240 @@
...
+# class TFIKagome(SpinSystemDataset):
+#     pass
+#
+#
+# class XXZChain(SpinSystemDataset):
+#     pass
+#
+#
+# class XXZLattice(SpinSystemDataset):
+#     pass
```
MichaelBroughton, line 240: It's probably best to not leave bulk comments in

MichaelBroughton:
@lamberta I did speak quickly with the TFDS folks and it seemed like a better idea that we stayed away from TFDS integration and just kept it local since we are farther away from "mainstream ml datasets" and have a pretty exotic dataformat in the form of `cirq.Circuit`s.

Our plan was to slowly grow out `tfq.datasets` and see where things go.

lamberta:
Gotcha. As far as hosting, it depends on the license. Some can be hosted from our GCP bucket, others need to be pulled from their source. cc: @yashk2810

yashk2810:
> hosted from our GCP bucket

AFAIK, we need an open license to host anything on our GCP bucket. As long as you have that, feel free to use it :)

tensorflow_quantum/datasets/spin_system.py
therooler: Doing this with a function should be fine.

But returning a circuit for each parameter, would'nt that lead to 80 copies of the same circuit? Or can we resolve multiple parameter sets for a single circuit? My reasoning was that it's best to create the circuit once, and then allow the user to feed the parameters in batches. Something like
```python
def tfi_chain(...):
    .
    .
    .
    return circuit, resolved_params, g, gs_energy, res_e
```

tensorflow_quantum/datasets/spin_system.py
therooler: No, probably not.

tensorflow_quantum/datasets/spin_system.py
therooler: That should be doable.

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,210 @@
...
+    def __init__(self, nspins: int, system_name: str):
+        """Instantiate this data set.
+
+        Args:
+            nspins: The number of spins in the system.
+            system_name: Name of the system we want to load.
+        """
+
+        # TODO: How should downloading and storing the data be handled?
+        data_path = '/tmp/data/' + system_name + "_{}/".format(int(nspins))
```
MarkDaoust, line 91: Prefer `tempfile.mkdtemp()`, it works on all platforms, "/tmp/" doesn't work on windows, for example.

Also note that temp directories are often cleared by reboots. Fo it you're expecting these to be big downloads you may want an option to set the save directory to somewhere durable, and/or a print out telling the iser where the data is.

therooler: The /tmp/ folder was a temporary solution (pun intended). I rewrote everything using `pathlib`. The default directry is now at `~/tfq-datasets` and the user can specify their own directory with an argument.

therooler:
So how should we handle the downloading? Can we use this GCP bucket for that?

MarkDaoust:
Yes, if the data is clearly open-licensed, then we can put it in one of our buckets (downloads.tensorflow.org, or TFQ could make their own.)

> 15 kb of data.

If that's the whole thing I'd be tempted to just put it right in the git-repository and distribute it directly with TFQ.

therooler:
I think adding the data to the repo is doable if you only store the circuit parameters, since this would be of the orders of Kb. However, if we want to add the exact ground states for larger systems, every data set will contain a 2^N complex vector in `.npy` format. That will probably be more than a couple of Kb. 

With regards to the open license, how do we arrange that? Do I need to sign something similar to the CLA?

karlunho:
@MichaelBroughton We have a process that we've put together for hosting datasets in GCP buckets. We use this for OpenFermion already.

tensorflow_quantum/datasets/spin_system.py
therooler: I implemented this by passing a function to each SpinSystem that creates a `cirq.PauliSum`  for depending on the order parameter `g`.

MarkDaoust:
> With regards to the open license, how do we arrange that? Do I need to sign something similar to the CLA?

If you're the author, I believe it's sufficient to just include a "LICENESE.txt" with the data download. But if it's going in the repo, then it should be covered by the CLA and top level "LICENSE.txt" file.

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    Transverse field Ising-model quantum data set
+
+    H = - \sum_{i} \sigma_i^z \sigma_{i+1}^z - g \sum_i \sigma_i^x
+
+    Contains 80 circuit parameterizations corresponding to
+    the ground states of the 1D TFI chain for g in [0.2,1.8].
+
+    Args:
+        qubits: cirq.Grid qubits of size (nspins,1)
```
MichaelBroughton, line 151: Nit: Python `lst` of `cirq.GridQubits` with length nspins

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,237 @@
...
+                g = float(str(directory.name))
+                lines = file.readlines()
+                res_e = float(lines[0].split('=')[1].strip('\n'))
+                fidelity = float(lines[0].split('=')[1].strip('\n'))
+            gs_energy = np.load(directory / Path('energy.npy'))[0]
+            parameters = np.load(directory / Path('params.npy'))
+            systems.append(
+                SpinSystem(g, res_e, fidelity, gs_energy, parameters, directory,
+                           self._hamiltonian_fn))
+        return sorted(systems)
```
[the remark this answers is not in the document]
therooler: To make sure that `SpinDataSet` is ordered according to the order parameter `g`.

therooler:
Hi,

Thanks for you comments! I simplified the interface using `namedtuple` and removed `SpinDataset` entirely, which allowed me to remove some other unnecessary code. The signature now matches the one you described. I also added some example code in the docstring that hopefully matches these formatting [requirements](https://www.tensorflow.org/community/contribute/docs_style). I ran the doc builder and checked out the Markdown files, which looked fine, but I think there may still be some errors there.

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,309 @@
...
+                SpinSystemInfo(g=g,
+                               gs=np.load(directory / Path('groundstate.npy')),
+                               gs_energy=np.load(directory /
+                                                 Path('energy.npy'))[0],
+                               res_energy=res_e,
+                               fidelity=fidelity,
+                               data_path=str(directory),
+                               params=np.load(directory / Path('params.npy'))))
+
+        return list(zip(*sorted(zip(order_parameters, additional_info))))
```
MichaelBroughton, line 103: Why does the returned dataset need to be sorted ? Does this break the ordering with the original circuits returned by tfi_chain ?

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,309 @@
...
+            circuit.append(
+                cirq.ZZ(qubits[j], qubits[j + 1])**(symbols[0, d] / np.pi))
+        for j in range(1, nspins - 1, 2):
+            circuit.append(
+                cirq.ZZ(qubits[j], qubits[j + 1])**(symbols[0, d] / np.pi))
+        if boundary_condition == 'closed':
+            circuit.append(
+                cirq.ZZ(qubits[nspins - 1], qubits[0])**(symbols[0, d] / np.pi))
+        for j in range(0, nspins):
+            circuit.append(cirq.rx(symbols[1, d])(qubits[j]))
```
MichaelBroughton, line 291: In this block it seems to be setting up a circuit with scalar multiples in the symbols only to be immediately resolved. We don't keep a reference copy of the unresolved circuit in additional info so as a power user I have no way of experimenting with the values of these parameters because I don't know what goes where.

Perhaps in additional info we should store an unresolved copy of the circuit along with cirq.ParamResolver that allow users to do this.

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,309 @@
...
+
+
+def tfi_chain(qubits, boundary_condition='closed', data_dir=None):
+    """Transverse field Ising-model quantum data set.
+
+    $$
+    H = - \sum_{i} \sigma_i^z \sigma_{i+1}^z - g \sum_i \sigma_i^x
+    $$
+
+    Contains 81 circuit parameterizations corresponding to
```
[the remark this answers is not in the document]
therooler: That sounds a lot better.

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,309 @@
...
+    the ground states of the 1D TFI chain for g in [0.2,1.8].
+
+    Example usage:
+
+    ```python
+    >>> import cirq
+    ... from tensorflow_quantum.datasets import tfi_chain
+    ... qbs = cirq.GridQubit.rect(4, 1)
+    ... circuit, labels, pauli_sums, addinfo  = tfi_chain(qbs, 'closed')
+    ```
```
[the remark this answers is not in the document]
therooler: Makes sense except for you final comment; if I run those lines in a Python interpreter I get a blank line.

tensorflow_quantum/datasets/spin_system.py
therooler: Great idea

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,309 @@
...
+
+    supported_n = [4, 8, 12, 16]
+    supported_bc = ['closed']
+
+    if not all([isinstance(q, cirq.GridQubit) for q in qubits]):
+        raise TypeError("qubits must be a list of cirq.GridQubit objects.")
+
+    nspins = len(qubits)
+
+    if not all([q.col == 0 for q in qubits]):
```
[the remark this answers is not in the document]
therooler: Actually we can :facepalm: , removed assertion

tensorflow_quantum/datasets/spin_system.py
therooler: Without the sort, `order_parameters` looks like this:
```python
[1.26, 1.7, 0.88, 1.22, 1.32, 0.9, 0.52, ...., 1.16, 0.5, 1.6, 0.4, 1.64, 0.38]
```
due to the way the folder are read. Sorting here does not break the resolved circuits from `tfi_chain`, since we loop over the sorted `order_parameters[i]` and `additional_info[i]`
```python
for i in range(len(order_parameters)):
    parameter_map = dict(
        zip(symbol_names.flatten(),
            additional_info[i].params.flatten()))
    resolved_circuits.append(cirq.resolve_parameters(
        circuit, parameter_map))
    hamiltonians.append(get_hamiltonian(qubits,
                                        order_parameters[i]))
```
I agree that it is maybe a bit hacky

therooler:
Getting closer. Thank you for all your help by the way @MichaelBroughton. Your comments have been really insightful to learn about writing user friendly code.

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,298 @@
...
+            circuit, parameter_map))
+
+        hamiltonians.append(get_hamiltonian(qubits, order_parameters[i]))
+
+    labels = np.zeros(len(order_parameters), dtype=np.int)
+    labels[np.array(order_parameters) < 1.0] = 0
+    labels[np.array(order_parameters) == 1.0] = 1
+    labels[np.array(order_parameters) > 1.0] = 2
+
+    return resolved_circuits, labels, hamiltonians, additional_info, circuit
```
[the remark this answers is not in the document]
therooler: The problem is that `SpinSystemInfo` is not system specific and immmutable, so adding a system dependent info to it is going to be ugly. Right now I changed the `params` field in `SpinSystemInfo` to be a dict that has three fields:

**Dict field created on initialization in `spin_system_data_set`**

- `raw`: Numpy float array of size (P,L) with the parameter values.

**Dict field added in `tfichain`**

- `symbol_names`: Numpy str array of size (P,L) with the symbol names.
- `param_resolver`: `cirq.ParamResolver` object containing the parameter to symbol mapping.

Usage:

```python
>>> qbs = cirq.GridQubit.rect(4,1)
>>> addinfo = tfi_chain(qbs)[3]
>>> addinfo[10].params['raw']
   [[2.39218603 2.1284263 ]
    [2.01284773 2.30447438]]
>>> addinfo[10].params['symbol_names']
   [['theta_0' 'theta_1']
    ['theta_2' 'theta_3']]
>>> addinfo[10].params['param_resolver']
   cirq.ParamResolver({'theta_0': 2.3921860302007274, 'theta_1': 2.128426301809212,
   'theta_2': 2.0128477257197104, 'theta_3': 2.304474378627067})
```
Which works, but is maybe too convoluted?

therooler:
I stored the data here: https://github.com/therooler/tfq_data
Let me know when you have the google storage thing set up, then I will integrate that into the keras downloader and test if everything works properly.

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,324 @@
...
+                exact diagonalization.
+            - `gs_energy`: Numpy `float` ground state energy from exact
+                diagonalization.
+            - `res_energy`: Python `float` residual between the circuit energy
+                and the exact energy from exact diagonalization.
+            - `fidelity`: Python `float` overlap between the circuit state
+                and the exact ground state from exact diagonalization.
+            - `params`: Dict with Python `str` keys and Numpy`float` values.
+                Contains $M \times P parameters. Here $M$ is the number of
+                parameters per circuit layer and $P$ the circuit depth.
```
yashk2810, line 80: It would be preferrable if you didn't indent these.

Keep it at the same level as `A Python lst of Python...`

Similar for all other Returns. I see this in other places too.

MichaelBroughton:
Big thanks to @yashk2810 for setting things up. Here's the link you can use in the code to download: https://storage.googleapis.com/download.tensorflow.org/data/quantum/spin_systems_data.zip

therooler:
Thanks @yashk2810! I got things to work with the link you supplied. However,  right now all the data is stored in one big .zip file. I think it might be better if each dataset  is zipped separately. Then if we add new data, the user does not have to download the whole thing again. Something like:

```
https://storage.googleapis.com/download.tensorflow.org/data/quantum/spin_systems/TFI_4_closed.zip
https://storage.googleapis.com/download.tensorflow.org/data/quantum/spin_systems/TFI_8_closed.zip
```

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,332 @@
...
+        raise TypeError("qubits must be a one-dimensional list")
+
+    if not all(isinstance(q, cirq.GridQubit) for q in qubits):
+        raise TypeError("qubits must be a list of cirq.GridQubit objects.")
+
+    nspins = len(qubits)
+
+    if nspins not in supported_n:
+        raise ValueError("Supported number of spins are {}, received {}".format(
+            supported_n, len(qubits)))
```
MichaelBroughton, line 275: len(qubits) -> nspins

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,332 @@
...
+
+    You can print the available order parameters
+
+    >>> [info.g for info in addinfo]
+    [0.20, 0.22, 0.24, ... ,1.76, 1.78, 1.8]
+
+    and the circuit corresponding to the ground state for a certain order
+    parameter
+
+    >>> print(circuits[10])
```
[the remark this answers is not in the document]
therooler: Only when you call print in Idle do you get the circuit representation. Same for `print(new_circuit)` below.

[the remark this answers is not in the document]
MichaelBroughton: Oops, my mistake.

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,332 @@
...
+    # Use Keras file downloader.
+    get_file(fname='spin_systems',
+             cache_dir=data_dir,
+             cache_subdir='',
+             origin="https://storage.googleapis.com/"
+             "download.tensorflow.org/data/"
+             "quantum/spin_systems_data.zip",
+             extract=True)
+    data_path = data_dir + "/tfq_data-master/spin_systems/" + system_name \
+                + "_{}".format(nspins)
```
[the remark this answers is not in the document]
therooler: This confused me a bit, the `get_file` code looks like this (line 215 to 224):
``` python
...
  if cache_dir is None:
    cache_dir = os.path.join(os.path.expanduser('~'), '.keras')
  if md5_hash is not None and file_hash is None:
    file_hash = md5_hash
    hash_algorithm = 'md5'
  datadir_base = os.path.expanduser(cache_dir)
  if not os.access(datadir_base, os.W_OK):
    datadir_base = os.path.join('/tmp', '.keras')
  datadir = os.path.join(datadir_base, cache_subdir)
  _makedirs_exist_ok(datadir)
...
```
So if the user passes path to `cache_dir` that does not exist, the data gets cached to `/tmp/.keras`, which is undesirable because then we store the data in a temporary folder. This is why I handled this with the `os.mkdir` block.

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,332 @@
...
+        if boundary_condition == "closed":
+            paulisum += -cirq.Z(qubits[0]) * cirq.Z(qubits[-1])
+        paulisum += -order_parameters[i] * sum(cirq.X(q) for q in qubits)
+        hamiltonians.append(paulisum)
+
+    # Set labels for the different phases.
+    labels = np.zeros(len(order_parameters), dtype=np.int)
+    labels[np.array(order_parameters) < 1.0] = 0
+    labels[np.array(order_parameters) == 1.0] = 1
+    labels[np.array(order_parameters) > 1.0] = 2
```
MichaelBroughton, line 325: In the docstring we say that `labels` is a Python `lst` but here it's actually a `np.ndarray`. I would opt to actually make it a list and construct it in the above for loop so that the output types are consistent with all the other python lists (even if they are going to convert everything over using `tfq.convert_to_tensor` or `tf.convert_to_tensor` anyway).

tensorflow_quantum/datasets/spin_system_test.py
```diff
@@ -0,0 +1,72 @@
...
+# ==============================================================================
+"""Test the spin system dataset"""
+import tensorflow as tf
+import numpy as np
+import cirq
+from tensorflow_quantum.datasets import spin_system
+
+
+class SpinSystemDataTest(tf.test.TestCase):
+    """Small test to make sure dataset for ClusterState works."""
```
MichaelBroughton, line 23: Testing TFI_chain

tensorflow_quantum/datasets/spin_system_test.py
```diff
@@ -0,0 +1,72 @@
...
+# limitations under the License.
+# ==============================================================================
+"""Test the spin system dataset"""
+import tensorflow as tf
+import numpy as np
+import cirq
+from tensorflow_quantum.datasets import spin_system
+
+
+class SpinSystemDataTest(tf.test.TestCase):
```
MichaelBroughton, line 22: Since it's possible that other SpinSystems may go into the `spin_system.py` module we might want to rename this class to indicate we are only testing the tfi_chain method and then make new classes for each of the new spin systems.

tensorflow_quantum/datasets/spin_system_test.py
```diff
@@ -0,0 +1,72 @@
...
+    def test_fidelity(self):
+        """Test that it returns the correct number of circuits."""
+        supported_nspins = [4, 8, 12, 16]
+        for nspins in supported_nspins:
+            qbs = cirq.GridQubit.rect(nspins, 1)
+            circuits, _, _, addinfo = spin_system.tfi_chain(
+                qbs,
+                'closed',
+            )
+            fidelities = []
```
MichaelBroughton, line 57: No need to store fidelities. You can just assert as you go without storing things.

tensorflow_quantum/datasets/spin_system_test.py
```diff
@@ -0,0 +1,72 @@
...
+            for n in range(len(addinfo)):
+                phi = cirq.Simulator().simulate(circuits[n]).final_state
+                gs = addinfo[n].gs
+                fidelities.append(np.abs(np.conj(np.dot(gs, phi))))
+            assert all([np.isclose(fid, 1.0, rtol=1e-3) for fid in fidelities])
```
MichaelBroughton, line 62: prefer self.Assert style of statement from this list: https://www.tensorflow.org/api_docs/python/tf/test/TestCase instead of vanilla assert statements.

tensorflow_quantum/datasets/spin_system_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def test_paulisum(self):
+        """Test that hamiltonian returns a PauliSum"""
+        qbs = cirq.GridQubit.rect(4, 1)
+        pauli_sums = spin_system.tfi_chain(qbs, 'closed')[2]
+        assert all(isinstance(ps, cirq.PauliSum) for ps in pauli_sums)
```
MichaelBroughton, line 68: Can we do a more involved test where you actually compute something using <psi | H | psi > ?

tensorflow_quantum/datasets/spin_system_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+
```
MichaelBroughton, line 70: Would be nice to have one or two more really basic tests to make sure we are returning the right number of circuits and that the param_resolvers work etc.

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,332 @@
...
+
+    Args:
+        qubits: Python `lst` of `cirq.GridQubit`s. Supported number of spins
+            are [4, 8, 12, 16].
+        boundary_condition: Python `str` indicating the boundary condition
+            of the chain. Supported boundary conditions are ["closed"]
+        data_dir: Optional Python `str` location where to store the data on
+            disk. Defaults to `/tmp/.keras`.
+    Returns:
+        A Python `lst` cirq.Circuit of depth N / 2 with resolved parameters.
```
MichaelBroughton, line 213: ... of `cirq.Circuit`s of depth `len(qubits)` / 2

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,332 @@
...
+        A Python `lst` of labels, 0, for the ferromagnetic phase (`g<1`), 1 for
+            the critical point (`g==1`) and 2 for the paramagnetic phase
+            (`g>1`).
+        A Python `lst` of `cirq.PauliSum`s
+        A Python `lst` of `namedtuple` instances containing the following
+            fields:
+        - `g`: Numpy `float` order parameter.
+        - `gs`: Complex numpy array ground state wave function from
```
MichaelBroughton, line 221: `np.ndarray`

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,332 @@
...
+        Python `str` of the form "theta_<integer>"
+
+    """
+    num = 0
+    while True:
+        yield "theta_" + str(num)
+        num += 1
+
+
+def download_spin_data(nspins, system_name, data_dir):
```
MichaelBroughton, line 58: Might want to change to `_download_spin_data` since we don't intend for this to be used outside of this module.

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,332 @@
...
+    labels[np.array(order_parameters) < 1.0] = 0
+    labels[np.array(order_parameters) == 1.0] = 1
+    labels[np.array(order_parameters) > 1.0] = 2
+
+    # Make sure that the data is ordered from small g=0.2 to g=1.8
+    _, resolved_circuits, labels, hamiltonians, additional_info = zip(*sorted(
+        zip(order_parameters, resolved_circuits, labels, hamiltonians,
+            additional_info)))
+
+    return resolved_circuits, labels, hamiltonians, additional_info
```
MichaelBroughton, line 332: Nit: newline at eof.

tensorflow_quantum/datasets/spin_system_test.py
therooler: I'm having some trouble using `cirq.unitary()` to construct the Hamilonian. Calling `cirq.unitary()` directly on the `PauliSum` does not work since the `PauliSum` does not have a `_unitary()` method. Calling it on the individual `PauliString` terms in the `PauliSum` works only if the coefficient of the `PauliString` has norm one.
line 326 in pauli_string.py:
```python
def _has_unitary_(self) -> bool:
    return abs(1 - abs(self.coefficient)) < 1e-6
```
so this fails:
```python
qbs = cirq.GridQubit.rect(4, 1)
pauli_sums = tfi_chain(qbs, 'closed')[2]
for ps in pauli_sums:
    for term in ps:
        print(term.coefficient)
        U = cirq.unitary(term)
```
```
(-1+0j)
(-1+0j)
(-1+0j)
(-1+0j)
(-0.2+0j)
Traceback (most recent call last):
  File "/home/rooler/Documents/tfq/tensorflow_quantum/datasets/testing_stuff.py", line 9, in <module>
    U = cirq.unitary(term)
  File "/home/rooler/anaconda3/envs/tfq/lib/python3.7/site-packages/cirq/protocols/unitary_protocol.py", line 151, in unitary
    "besides None or NotImplemented.".format(type(val), val))
TypeError: cirq.unitary failed. Value doesn't have a (non-parameterized) unitary effect.

type: <class 'cirq.ops.pauli_string.PauliString'>
value: ((-0.2+0j)*cirq.X(cirq.GridQubit(0, 0)))

The value failed to satisfy any of the following criteria:
- A `_unitary_(self)` method that returned a value besides None or NotImplemented.
- A `_decompose_(self)` method that returned a list of unitary operations.
- An `_apply_unitary_(self, args) method that returned a value besides None or NotImplemented.
```
Also, `cirq.unitary()` returns only local unitaries, not the unitary with respect to the full Hilbert space. So the ZZ terms here are just 4x4 matrices instead of 2^4 x 2^4 (for 4 qubits). This would require you to tensor the individual unitaries based on the qubit number yourself.

MichaelBroughton: There is an open issue in Cirq for adding a `.matrix()` function to PauliSum objects. You can do expectation much more compactly with something like:
```python
my_paulisum.expectation_from_wavefunction(numpy_state_vector)
```

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -0,0 +1,317 @@
...
+
+    # Use Keras file downloader.
+    tf.keras.utils.get_file(fname='spin_systems',
+                            cache_dir=data_dir,
+                            cache_subdir='',
+                            origin="https://storage.googleapis.com/"
+                            "download.tensorflow.org/data/"
+                            "quantum/spin_systems_data.zip",
+                            extract=True)
+    data_path = data_dir + "/tfq_data-master/spin_systems/" + system_name \
```
MichaelBroughton, line 85: Prefer `os.path.join` like used below.

tensorflow_quantum/datasets/spin_system_test.py
```diff
@@ -0,0 +1,97 @@
...
+            qbs = cirq.GridQubit.rect(nspins, 1)
+            circuits, _, pauli_sums, addinfo = spin_system.tfi_chain(
+                qbs, 'closed')
+            qubit_map = {qbs[i]: i for i in range(len(qbs))}
+            for n in range(len(pauli_sums)):
+                phi = cirq.Simulator().simulate(circuits[n]).final_state
+                e = pauli_sums[n].expectation_from_wavefunction(phi, qubit_map)
+                self.assertAllClose(e, addinfo[n].gs_energy, rtol=1e-4)
+
+    def test_returned_objects(self):
```
MichaelBroughton, line 78: Awesome! One final test that would be nice to see is that if I take a circuit from additional_info and resolve in the paramresolver do I get the original datapoint circuit ? This could be done via comparing the final wavefunctions for approximate equality.

therooler:
Wonderful!

There is one final thing, I feel like the folder structure for the data is kinda of ugly right now.
The .zip we download has structure `tfq_data-master/spin_systems/TFI_closed_x`. I think
restructuring the data as follows will be worth it in the long run, if we plan on adding more systems:

At `https://storage.googleapis.com/download.tensorflow.org/data/quantum/spin_systems/`
| **system name** | **boundary conditions** | **system size** | **order parameter**|
|--------------|------------|--------------|-------------|
| `TFI_chain/`      |   `closed/`         |       `4/`       | `0.2/` |
| `XXZ_chain/`    |     `open/`       |           `8/`      | `0.22/` | 
| `TFI_rectangular/` |       `torus/`          |   `12/`             | `0.40_1.80/`| 
|       ...              ` |       `cylinder/`          |      `3x3/`          |`0.42_1.80/` | 
|--------------|------------|--------------|---------|

This structure would give more freedom when adding more different kinds of spin systems, but keeps things clear imo.

We could zip at the level of `system_name` or `boundary_conditons`, to limit the file size. To test this, I temporarily hosted the `TFI_chain` data in a git [repo,](https://github.com/therooler/TFI_chain/) so that it passes the CI stuff right now. 

Let me know if you agree.

MichaelBroughton:
Great idea, let's zip on `system_name`. I've spoken with @yashk2810 and he's got us all setup with 
https://storage.googleapis.com/download.tensorflow.org/data/quantum/spin_systems/TFI_chain.zip now. Feel free to hook things in and ping me here once you've got it working and we can merge.

therooler:
Should be good to go!

PR #223, opened 2020-05-01
Make GateFused a proper class

### LINKED ISSUE #67: Push GateFused implementations into cc file

zaqqwerty:
Resolves #67

### DIFF
--- a/tensorflow_quantum/core/qsim/fuser_basic.cc
+++ b/tensorflow_quantum/core/qsim/fuser_basic.cc
@@ -54,8 +54,10 @@ void GateFused::GetAllGates(const std::vector<const Gate*>** ref) const { *ref =
 
 unsigned int GateFused::GetNumGates() const { return gates_.size(); }
 
-const Gate* GateFused::GetGate(unsigned int gate_index) const {
-  return gates_.at(gate_index);
+// Return a copy of the gate pointed to by the given index
+const Gate GateFused::GetGate(unsigned int gate_index) const {
+  Gate gate_copy(*gates_.at(gate_index));
+  return gate_copy;
 }
 
 void GateFused::SetGate(unsigned int gate_index, const Gate* gate) {
@@ -91,7 +93,7 @@ bool operator==(const GateFused& l, const GateFused& r) {
     return false;
   }
   for (size_t i = 0; i < l.GetNumGates(); i++) {
-    if (*l.GetGate(i) != *r.GetGate(i)) {
+    if (l.GetGate(i) != r.GetGate(i)) {
       return false;
     }
   }
--- a/tensorflow_quantum/core/qsim/fuser_basic.cc
+++ b/tensorflow_quantum/core/qsim/fuser_basic.cc
@@ -31,16 +31,74 @@ using ::tensorflow::Status;
 // Appends to `fused_gates` all single-qubit gates on the current `qubit_wire`,
 // until reaching the end of the wire or a multi-qubit gate.
 // Starts the search at `current_timeslice`.
-void Advance(const std::vector<const Gate*>& qubit_wire,
-             std::vector<const Gate*>* fused_gates,
+void Advance(const std::vector<const Gate*>& qubit_wire, GateFused* gate,
              unsigned int* current_timeslice) {
   while (*current_timeslice < qubit_wire.size() &&
          qubit_wire[*current_timeslice]->num_qubits == 1) {
-    fused_gates->push_back(qubit_wire[(*current_timeslice)++]);
+    gate->AddGate(qubit_wire[(*current_timeslice)++]);
   }
 }
 
 }  // namespace
+
+GateFused::GateFused(const unsigned int time, const unsigned int q0,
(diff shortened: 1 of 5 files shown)


tensorflow_quantum/core/qsim/fuser_basic.cc
```diff
@@ -31,17 +31,75 @@ using ::tensorflow::Status;
...
+GateFused::GateFused(const unsigned int time, const unsigned int q0,
+                     const unsigned int q1, const Gate* anchor)
+    : time_(time), anchor_(anchor) {
+  qubits_[0] = q0;
+  qubits_[1] = q1;
+}
+
+void GateFused::AddGate(const Gate* gate) { gates_.push_back(gate); }
+
+std::vector<const Gate*> GateFused::GetAllGates() const { return gates_; }
```
MichaelBroughton, line 53: Does this return the object that contains the pointers, adding some unnecessary overhead ? It might be better to do something like 
```C++
GetAllGates(std::vector<const Gate*>* ref) { ref = &gates_; }
```

tensorflow_quantum/core/qsim/fuser_basic.cc
```diff
@@ -31,17 +31,75 @@ using ::tensorflow::Status;
...
+
+unsigned int GateFused::GetNumGates() const { return gates_.size(); }
+
+const Gate* GateFused::GetGate(unsigned int gate_index) const {
```
MichaelBroughton, line 57: Not as big a deal as above, but would be safer to do an assignment instead of returning a pointer.

tensorflow_quantum/core/qsim/fuser_basic.cc
```diff
@@ -31,17 +31,75 @@ using ::tensorflow::Status;
...
   while (*current_timeslice < qubit_wire.size() &&
          qubit_wire[*current_timeslice]->num_qubits == 1) {
-    fused_gates->push_back(qubit_wire[(*current_timeslice)++]);
+    gate->AddGate(qubit_wire[(*current_timeslice)++]);
   }
 }
 
 }  // namespace
 
+GateFused::GateFused(const unsigned int time, const unsigned int q0,
```
MichaelBroughton, line 44: should it be `const Gate& anchor` instead ? it looks like you are passing in `&a_gate` in a lot of the tests

zaqqwerty: `&a_gate` passes in a pointer, while `const Gate& anchor` expects a reference, right?  I don't think a reference would work here because we are storing a way to look at the anchor gate again later, rather than mutating an external object

tensorflow_quantum/core/qsim/fuser_basic.cc
zaqqwerty: Switched to parameter output

tensorflow_quantum/core/qsim/fuser_basic.cc
zaqqwerty: Now returns copy of the pointed-to gate instead of a pointer

tensorflow_quantum/core/qsim/fuser_basic.cc
```diff
@@ -31,17 +31,77 @@ using ::tensorflow::Status;
...
+                     const unsigned int q1, const Gate* anchor)
+    : time_(time), anchor_(anchor) {
+  qubits_[0] = q0;
+  qubits_[1] = q1;
+}
+
+void GateFused::GetAllGates(const std::vector<const Gate*>** ref) const {
```
MichaelBroughton, line 51: Any particular reason you went for a double pointer here ?

tensorflow_quantum/core/qsim/fuser_basic.cc
MichaelBroughton: Sorry, that wasn't what I had in mind, returning the full object is definitely unnecessary. Let's just stick with the pointer.

zaqqwerty: Reverted to pointer return

tensorflow_quantum/core/qsim/fuser_basic.cc
zaqqwerty: No, now switched to pointer reference and it works too

zaqqwerty:
1) It would be better not to reference object parameters, but we already do this heavily with our `Gate*` pointers to things stored in `Circuit`s.  In the future we can hopefully get rid of these practices.
2) Inlined `AddGate`
3) Ran ```bazel test -c opt --cxxopt="-D_GLIBCXX_USE_CXX11_ABI=0" --cxxopt="-msse2" --cxxopt="-msse3" --cxxopt="-msse4" --cxxopt="-fsanitize=address" --linkopt="-fsanitize=address" --cxxopt="-g" --cxxopt="-O0" //tensorflow_quantum/core/qsim:all``` and everything passed

zaqqwerty:
Not a priority now

PR #225, opened 2020-05-04
Fixed Cirq version problems.

MichaelBroughton:
Cirq 0.8 currently breaks us. Fixing `setup.py` to hard pin to 0.7. Let me know if I've missed anything

### DIFF
--- a/release/setup.py
+++ b/release/setup.py
@@ -50,7 +50,7 @@ class InstallPlatlib(install):
             self.install_lib = self.install_platlib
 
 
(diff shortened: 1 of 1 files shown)


MichaelBroughton:
Ahhh good point, it might make sense to add a pip install cirq ==0.7.0 into the notebooks in the meantime. This definitly makes our next release a little more urgent. I'll take a look at how we're doing and see if it makes sense for us to cut a new release in the coming weeks too.

PR #234, opened 2020-05-06
Bumped Version to cirq 0.8.

### LINKED ISSUE #228: Upgrade to cirq 0.8
Cirq 0.8 is out. Lets see if we can upgrade to depending on it.

MichaelBroughton:
Fixes #228


MichaelBroughton:
Shouldn't we do that when we cut 0.3 ?

lamberta:
Notebook seems to run so I guess it's fine: https://colab.research.google.com/github/tensorflow/quantum/blob/cirq_08/docs/tutorials/barren_plateaus.ipynb
I was just a little surprised when the PR title mentioned the bump to 0.8 and the notebook is still using 0.7 🤷‍♂️

PR #236, opened 2020-05-08
Precision fix

### LINKED ISSUE #235: State simulator dfference between Cirq and TFQ
Environment:
**Python 3.7**
**tensorflow = 2.1.0**
**tensorflow-quantum = 0.3.0**
**Cirq = 0.7.0**

**Issue:** When running larger (N = 12) variational circuits `cirq.Simulator()` and `tfq.layer.State` output states that do not agree. When using this in VQE to find the ground state of some Hamiltonian, after convergence, the Cirq state has overlap ~1 with the ground state whereas the TFQ state does not. This problem gets worse as N increases.

**Reproducible example**
We run a depth 4 circuit with ZZ gates between nearest neighbors on an open chain of length N.
With N=8, the overlap between the two states is 1. But when we go to a chain of length 12, this is no longer the case.

```python
import cirq
import tensorflow as tf
import numpy as np
import sympy


def unique_name():
    """
    Generator to generator an infinite number of unique names

    Yields: Strings of the form 'theta_<number>'

    """
    num = 1
    while True:
        yield 'theta_' + str(num)
        num += 1


def issue(nqubits):
    depth = 4

    name_generator = unique_name()

    qubits = cirq.GridQubit.rect(nqubits, 1)
    circuit = cirq.Circuit()

    control_params = []

    circuit.append(cirq.H.on_each(qubits))
    for _ in range(depth):
        name = next(name_generator)
        control_params.append(name)
        symbol = sympy.Symbol(name)

        circuit.append(cirq.ZZ(qubits[i], qubits[i + 1]) ** symbol for i in
                       range(0, nqubits - 1, 2))
        circuit.append(cirq.ZZ(qubits[i], qubits[i + 1]) ** symbol for i in
                       range(1, nqubits - 1, 2))
        circuit.append(cirq.ZZ(qubits[nqubits - 1], qubits[0]) ** symbol)

    initial_value = np.random.rand(1, len(control_params))

    params_dict = dict(zip(control_params, initial_value[0]))
    resolver = cirq.ParamResolver(params_dict)
    phi_cirq = cirq.Simulator().simulate(circuit, resolver).final_state

    state_layer = tfq.layers.State()
    model_params = tf.Variable(initial_value=initial_value)
    # phi_tfq = \
    #     state_layer(circuit, symbol_names=control_params,
    #                 symbol_values=initial_value)[
    #         0].numpy()
    phi_tfq = \
        state_layer(circuit, symbol_names=control_params,
                    symbol_values=model_params)[
            0].numpy()

    overlap = np.abs(np.conj(phi_cirq) @ phi_tfq)
    return overlap


if __name__ == "__main__":
    print(issue(8))
    print(issue(12))

>>> 0.9999997
>>> 0.5101427
```
Is our implementation incorrect? Or have we encountered a bug?

MichaelBroughton:
Fixes #235

### DIFF
--- a/tensorflow_quantum/core/ops/circuit_execution_ops_test.py
+++ b/tensorflow_quantum/core/ops/circuit_execution_ops_test.py
@@ -60,175 +60,178 @@ SIMS = [WF_SIM, WF_SIM, DM_SIM]
 class OpGetterInputChecks(tf.test.TestCase):
     """Check that the op getters handle inputs correctly."""
 
-#     def test_get_expectation_inputs(self):
-#         """Test that get expectation only accepts inputs it should."""
-#         circuit_execution_ops.get_expectation_op()
-#         circuit_execution_ops.get_expectation_op(backend=cirq.Simulator())
-#         circuit_execution_ops.get_expectation_op(
-#             backend=cirq.DensityMatrixSimulator())
-#         circuit_execution_ops.get_expectation_op()
-#         with self.assertRaisesRegex(NotImplementedError,
-#                                     expected_regex='Sample-based'):
-#             mock_engine = mock.Mock()
-#             circuit_execution_ops.get_expectation_op(
-#                 cirq.google.QuantumEngineSampler(engine=mock_engine,
-#                                                  processor_id='test',
-#                                                  gate_set=cirq.google.XMON))
-#         with self.assertRaisesRegex(
-#                 TypeError, expected_regex="a Cirq.SimulatesFinalState"):
-#             circuit_execution_ops.get_expectation_op(backend="junk")
+    def test_get_expectation_inputs(self):
+        """Test that get expectation only accepts inputs it should."""
+        circuit_execution_ops.get_expectation_op()
+        circuit_execution_ops.get_expectation_op(backend=cirq.Simulator())
+        circuit_execution_ops.get_expectation_op(
+            backend=cirq.DensityMatrixSimulator())
+        circuit_execution_ops.get_expectation_op()
+        with self.assertRaisesRegex(NotImplementedError,
+                                    expected_regex='Sample-based'):
+            mock_engine = mock.Mock()
+            circuit_execution_ops.get_expectation_op(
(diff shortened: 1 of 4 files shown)


tensorflow_quantum/core/ops/circuit_execution_ops_test.py
```diff
@@ -224,6 +224,35 @@ def test_simulate_state_with_symbols(self, op_and_sim, n_qubits,
...
+                                   for (op,
+                                        sim) in zip(STATE_OPS[:-1], SIMS[:-1])],
+                })))
+    def test_simulate_state_large(self, op_and_sim):
+        """Test a reasonably large and complex circuit."""
+        op, sim = op_and_sim
+        symbol_names = []
+        circuit_batch, resolver_batch = \
+            util.random_circuit_resolver_batch(
+                cirq.GridQubit.rect(4, 4), 5)
```
zaqqwerty, line 241: I think we also need to test the specific case from the original bug, which used a line of 12 qubits

tensorflow_quantum/core/src/program_resolution_test.cc
```diff
@@ -84,6 +114,18 @@ TEST(ProgramResolutionTest, ResolveQubitIdsInvalidArg) {
             tensorflow::Status(
                 tensorflow::error::INVALID_ARGUMENT,
                 "Found a Pauli sum operating on qubits not found in circuit."));
+
+  ASSERT_TRUE(
```
zaqqwerty, line 118: Add a test for first index being junk

tensorflow_quantum/core/ops/circuit_execution_ops_test.py
MichaelBroughton: The case of (4,4) actually proved to be worse than the (1,12) case in my testing.

tensorflow_quantum/core/src/program_resolution_test.cc
MichaelBroughton: Fixed.

PR #237, opened 2020-05-08
Tutorial mistake fix: for test examples, use len(x_test)

yuanoook:


yuanoook:
@googlebot I signed it!

PR #239, opened 2020-05-11
Prepare for 0.3.

MichaelBroughton:
Going to cut 0.3 right after which will auto install cirq 0.8 now


lamberta:
Update/remove Cirq from [install.md](https://github.com/tensorflow/quantum/blob/master/docs/install.md) ?

MichaelBroughton:
Ahhh good call. I updated th install instructions. PTAL

PR #241, opened 2020-05-14
Implement rotosolve minimizer

SachinCompton:
Here we implement the rotosolve minimizer for tensorflow quantum.


SachinCompton:
> @googlebot I signed it!

@googlebot I signed it!

PR #243, opened 2020-05-18
Add unsigned quantum integer tools

zaqqwerty:
Added utility functions for treating groups of qubits as unsigned quantum integers.


zaqqwerty:
I see what you meant about the possible obscurity of the interface.  I took out more of the overly stiff code.  I left in a function to take position operators to momentum operators, and also a function to get sets of qubits as registers; do things seem better now as library code?

tensorflow_quantum/python/util.py
```diff
@@ -519,3 +519,95 @@ def exponential(operators, coefficients=None):
...
             circuit += cirq.Circuit(exp_circuit)
     return circuit
+
+
+# ==============================================================================
+# Unsigned quantum integer functions.
+# ==============================================================================
+
+
+def projector_on_one(qubit):
```
MichaelBroughton, line 529: Don't think this is complicated enough to where it needs it's own function.

tensorflow_quantum/python/util.py
```diff
@@ -519,3 +519,95 @@ def exponential(operators, coefficients=None):
...
+
+    Returns:
+        `cirq.PauliSum` representing the projector.
+    """
+    if not isinstance(qubit, cirq.GridQubit):
+        raise TypeError("A projector must live on a cirq.GridQubit.")
+    return 0.5 * cirq.I(qubit) - 0.5 * cirq.Z(qubit)
+
+
+def registers_from_precisions(precisions):
```
MichaelBroughton, line 546: This is still opaque and hard for a user to build on. Suppose I call this function then want to apply some gates to the qubits that correspond to the register that goes with precisions[3]. How do I figure out which qubits those are without reading the source code of this function (even when calling `to_momentum_basis` I get stuck)?

tensorflow_quantum/python/util.py
```diff
@@ -519,3 +519,95 @@ def exponential(operators, coefficients=None):
...
+    Args:
+        qubits: a Python `list` of qubits on which the circuit is supported.
+        circuit: a `cirq.Circuit` to be taken to the momentum basis.
+
+    Returns:
+        `cirq.Circuit` which is the momentum basis version of the given circuit.
+    """
+    convert = cirq.ConvertToCzAndSingleGates(allow_partial_czs=True)
+    transform = cirq.Circuit(cirq.QFT(*register))
+    convert(transform)
```
MichaelBroughton, line 612: Can this just be `cirq.decompose` instead ?

zaqqwerty: cirq.decompose left gates we do not support

zaqqwerty:
I want to add support for the Somma convention X and P operators eventually too, which I was thinking would operate on registers.  The code in the version 1 code base basically defines generalized PauliString and PauliSum objects which contain general operator tensor products and linear combinations, not just IXYZ objects.  For example, the P operators had, instead of `to_z_basis_ops`, a convert to X basis circuit for each register.  X and P and these integer position momentum ops would be special cases of that framework

tensorflow_quantum/python/util.py
zaqqwerty: Good point, removed this

tensorflow_quantum/python/util.py
zaqqwerty: This is a function for making a ragged tensor of qubits, which is why the`GridQubit.rect` functionality won't work.  Also, results are meant to be consumed only by functions expecting lists of qubits which are unique, not functions for which individual qubits are special.  For me the idea of designating a set of qubits as a register is that none are supposed be addressed individually by user code

MichaelBroughton: > For me the idea of designating a set of qubits as a register is that none are supposed be addressed individually by user code.

I agree. suppose I want to do a QFT on a register. How do I hope to know which qubits correspond to which register without knowing this particular row and column pattern for how this code allocated registers. We don’t want library code having to bake in this assumption.

MichaelBroughton:
That seems like a reasonably sized contribution. Could we write a design document outlining this implementation ? It feels like fitting the idea of different  register formats, understanding how these things fit in with optimizers, API design etc isn’t so straightforward.

zaqqwerty:
Closing for now

PR #246, opened 2020-05-26
Updated two_qubit_unitary function

vinitX:
The two-qubit unitary function should contain 15 parameters but only 14 were active due to an error in the parameter assignment of the coupling gates. They should be parameterized with symbols 6,7,8 and not 7,8,9.


vinitX:
> @googlebot I signed it!

PR #258, opened 2020-06-08
Added msan script and CI test.

MichaelBroughton:
Testing the raw C++ files that don't have any python or TF boilerplate. This does exclude our `ops`, but it's better than nothing.


MichaelBroughton:
PTAL

PR #270, opened 2020-06-18
Function factory

SachinCompton:
Merge function factory to master branch.


tensorflow_quantum/python/optimizers/utils.py
```diff
@@ -0,0 +1,104 @@
...
+from __future__ import print_function
+from operator import mul
+from functools import reduce
+import tensorflow as tf
+
+
+def function_factory(model, loss, train_x, train_y):
+    """A factory to create a function required by tfq.optimizer.rotosolve.
+
+    This function is originally defined for l-bgfs minimizer for tensorflow
```
MichaelBroughton, line 28: This seems like you are providing a more general feature for the library that something that is just for rotosolve. If this is copied from l-bfgs can you provide a link to it ? If it is similar enough perhaps we don't need to write it ourselves and can instead depend on the implementation from tf.

SachinCompton: This function is introduced from internet https://pychao.com/2019/11/02/optimize-tensorflow-keras-models-with-l-bfgs-from-tensorflow-probability/ but not inside the TensorFlow probability package.

SachinCompton:
Thanks Michael, let me figure out what's the best way to implement it.

PR #294, opened 2020-07-11
Deserializer rounding

jaeyoo:
This PR fixes precision issue during serialization & deserialization.

### DIFF
--- a/tensorflow_quantum/core/serialize/serializer.py
+++ b/tensorflow_quantum/core/serialize/serializer.py
@@ -17,6 +17,7 @@
 import copy
 import numbers
 import sympy
+import numpy as np
 
 import cirq
 import cirq.google.api.v2 as v2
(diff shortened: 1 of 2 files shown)


jaeyoo:
> Feel free to merge. Can we up this to 7 or 8 or does it become unstable at that point ?

No we can't. It shows discrepancy at 7 or 8. Thanks.

tensorflow_quantum/core/serialize/serializer.py
```diff
@@ -26,6 +27,10 @@
 _CONSTANT_TRUE = lambda x: True
 
 
+def _round(x):
+    return np.round(x, 6) if isinstance(x, float) else x
```
MichaelBroughton, line 31: Are we ok with the fact that:
```python3
isinstance(np.float32(0.1), float) => False
```
 ?

jaeyoo: hmm we don't use `np.float32`. if we use, even if the value itself is exactly the same, circuit equality failed due to types.

jaeyoo: so it is intended.

jaeyoo: The reason why I failed to update serializer was using `np.float32`. I realized that `np.round` itself doesn't require the data type, so I could do it today :)

Issue #263, opened 2020-06-14
[Tracking] qsim migration


MichaelBroughton:
Now that qsim has open sourced and has enough features and tests, I think it's time we move to depending on them and having one version of qsim instead of ours and theirs. In order to begin the changeover this will roughly entail:

1. ~renaming our `qsim` namespace to `qsim_old` to avoid collisions and make development easier.~
2. ~Add a `circuit_parser_qsim.h` file in `src` that ingests `cirq::api::google::v2::Program` protos and converts them to a form compatible with: https://github.com/quantumlib/qsim/blob/master/lib/circuit.h~
3. ~Wait for qsim to implement a publicly visible `FOR` interface that we can hook into tensorflow parallelization. Neither `seqfor` and `parfor` will work for us ( we should also likely make the switch to the new parallelization scheme in: #255 ).~
4. ~Modify implementation of `tfq_simulate_state_op.cc` to use new qsim dependencies.~
5. ~Modify implementation of `tfq_simulate_expectation_op.cc`.~
6. ~Modify implementation of `tfq_simulate_sampled_expectation_op.cc`~
7. ~Modify implementation of `tfq_calculate_unitary_op.cc`~
8. Remove `qsim` directory from our source.

Once we get this done it should be easy to add new gate support since their qsim now has a gateset comparable to ours.

CC: @jaeyoo @95-martin-orion @zaqqwerty @

zaqqwerty:
Tagging #171, TBD if this is needed or if external qsim handles this well

MichaelBroughton:
External qsim has since fixed their fuser.

MichaelBroughton:
Now that we are on Step 4 and want to integrate qsim into our first op which @jaeyoo wants to take on, we need to do the following things:

1. Implement our own TFQFor to use with qsim that has the same Run and RunReduce function as this: https://github.com/quantumlib/qsim/blob/master/lib/parfor.h . The reason we need this is because openMP isn't supported in TF and we need to use TF threading: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/threadpool.h (note this gets exported as tensorflow/core/lib/core/threadpool.h).

2. Rewrite the code in `tfq_simulate_op.cc` to use our new `circuit_parser_qsim`.

As a final note we want to change parallelization mechanisms to parallelizing in the same fashion as OSS qsim does now (see #255 ). This should make the rewrite of `tfq_simulate_op.cc` a lot simpler.

[...]

PR #300, opened 2020-07-14
Ported Z basis parsing to OSS qsim.

MichaelBroughton:
Porting over ZBasis conversion code from `circuit_parser` to `circuit_parser_qsim` in the ongoing effort here: #263 . This brings us one step closer to getting tfq_sampled_expectation onto OSS qsim too.

### DIFF
--- a/tensorflow_quantum/core/src/circuit_parser_qsim.cc
+++ b/tensorflow_quantum/core/src/circuit_parser_qsim.cc
@@ -458,7 +458,7 @@ Status QsimZBasisCircuitFromPauliTerm(
     transform_exponent = -0.5;
     gate_type = "Y";
     if (pair.pauli_type() == "Y") {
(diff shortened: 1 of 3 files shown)


tensorflow_quantum/core/src/circuit_parser_qsim.cc
```diff
@@ -438,4 +438,48 @@ Status QsimCircuitFromPauliTerm(
...
+    if (pair.pauli_type() == "Z") {
+      // Z requires no transform.
+      continue;
+    }
+
+    // Assume it is X and the transform is Y^-0.5
+    transform_exponent = -0.5;
+    gate_type = "Y";
+    if (pair.pauli_type() == "Y") {
+      // Y regquires X**0.5 transform.
```
jaeyoo, line 461: requires

PR #307, opened 2020-07-16
Smarter parallelization of state op

MichaelBroughton:
Noticed that there are certain cases where parallelizing over wavefunctions vs parallelizing over batches have their benefits. TL;DR Circuit smaller than 26 qubits can parallelize over the batch. Circuits larger than 26 qubits should parallelize over wavefunctions and process the batch serially. 

More specifically:
If the circuits are less than ~26 qubits , most machines (https://cloud.google.com/compute/docs/machine-types#e2_machine_types) should have enough memory to simulate circuits entirely isolated in their own threads. This is nice because we can assign large workloads to each thread without having to worry about memory or whether or not the threads have enough work to do.

If the circuits are more than 26 qubits we then run the risk of this O(num_threads * size of state vector) memory usage becoming a pretty serious problem so we can move to parallelize over a single wave function instead. Once it is that large, this should keep the CPU busy long enough to benefit from the alternate parallelization scheme.

### DIFF
--- a/tensorflow_quantum/core/ops/tfq_simulate_state_op.cc
+++ b/tensorflow_quantum/core/ops/tfq_simulate_state_op.cc
@@ -18,6 +18,7 @@ limitations under the License.
 #include "../qsim/lib/circuit.h"
 #include "../qsim/lib/gate_appl.h"
 #include "../qsim/lib/gates_cirq.h"
+#include "../qsim/lib/seqfor.h"
 #include "../qsim/lib/simmux.h"
 #include "cirq/google/api/v2/program.pb.h"
 #include "tensorflow/core/framework/op_kernel.h"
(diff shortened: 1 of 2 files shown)


tensorflow_quantum/core/ops/tfq_simulate_state_op.cc
```diff
@@ -131,23 +159,62 @@ class TfqSimulateStateOp : public tensorflow::OpKernel {
...
+        }
+        for (uint64_t j = (uint64_t(1) << nq);
+             j < (uint64_t(1) << max_num_qubits); j++) {
+          (*output_tensor)(i, j) = std::complex<float>(-2, 0);
+        }
+      }
+      sv.release();
+    };
+
+    const int num_cycles = 200 * (1 << max_num_qubits);
```
zaqqwerty, line 215: Wondering why this number is different than at https://github.com/tensorflow/quantum/blob/1bccab0e8751fa062b728ab150ac15420217a881/tensorflow_quantum/core/src/util_qsim.h#L56 ?

tensorflow_quantum/core/ops/tfq_simulate_state_op.cc
```diff
@@ -91,13 +92,40 @@ class TfqSimulateStateOp : public tensorflow::OpKernel {
...
+
+ private:
+  void ComputeLarge(
+      const std::vector<int>& num_qubits, const int max_num_qubits,
+      const std::vector<std::vector<qsim::GateFused<QsimGate>>>& fused_circuits,
+      tensorflow::OpKernelContext* context,
+      tensorflow::TTypes<std::complex<float>, 1, long int>::Matrix*
+          output_tensor) {
     // Instantiate qsim objects.
     const auto tfq_for = tfq::QsimFor(context);
```
zaqqwerty, line 127: To be sure I understand, is it correct to say that, given `tfq::QsimFor`, the `qsim::Simulator` will use multiple threads managed by TF to update many parts of the same wavefunction in parallel upon gate application?  In contrast to the parallelization in `ComputeSmall`, where each TF thread manages its own wavefunction, and in each thread a QSim simulator applies gates to only one site in that wavefunction at a time (since `qsim::SequentialFor` is given a thread count of 1 to work with inside the TF thread it is called from)?

tensorflow_quantum/core/ops/tfq_simulate_state_op.cc
MichaelBroughton: For this simulation technique the unit of work is "the cost of simulating one circuit end to end and computing the expectation value" for the other the unit of work is "the time to update ONE amplitude entry associated with a certain index". One scales with system size while the other is a constant.

tensorflow_quantum/core/ops/tfq_simulate_state_op.cc
MichaelBroughton: That is correct.

MichaelBroughton:
OK, turns out the OS in CI just doesn't have enough memory for this to work (Doesn't look like there is an easy fix for instance upgrades either). Should we just go ahead and leave this uncovered ? I did test both version locally.

zaqqwerty:
Sounds good

PR #319, opened 2020-07-22
299 parameter shift exposure

zaqqwerty:
Exposes the parse_programs util function and impure symbol name


zaqqwerty:
Going to open a PR with a different solution soon

Issue #286, opened 2020-07-06
Circuit parameter filler op and layer


zaqqwerty:
It seems a useful and general capability in TFQ would be a user facing parameter resolver.  Instead of:
```python
# Build a circuit to prepare computational basis states
qubits = cirq.GridQubit.rect(1, 4)
circuit = cirq.Circuit()
symbols = []
for n, q in enumerate(qubits):
    new_bit = sympy.Symbol("_bit_{}".format(n))
    circuit += cirq.X(q) ** new_bit
    symbols.append(new_bit)

# Resolve circuit for a list of basis states
bitstring_list = [[0, 0, 0, 0], [0, 1, 0, 1], [1, 0, 1, 1]]
resolved_circuits = []
for bitstring in bitstring_list:
    resolve_dict = {}
    for s, b in zip(symbols, bitstring):
        resolve_dict[s] = b
    resolved_circuits.append(cirq.resolve_parameters(circuit, cirq.ParamResolver(resolve_dict)))
```
I would like to be able to generate the list of bitstrings via a TensorFlow op, so that I would need to resolve the circuits via something like:
```python
# Build a circuit to prepare computational basis states
qubits = cirq.GridQubit.rect(1, 4)
circuit = cirq.Circuit()
symbols = []
for n, q in enumerate(qubits):
    new_bit = sympy.Symbol("_bit_{}".format(n))
    circuit += cirq.X(q) ** new_bit
    symbols.append(new_bit)

# Resolve circuit for a tensor of basis states
bitstring_tensor = tf.constant([[0, 0, 0, 0], [0, 1, 0, 1], [1, 0, 1, 1]])
symbols_tensor = tf.convert_to_tensor([str(s) for s in symbols])
circuit_tensor = tfq.convert_to_tensor([circuit])
resolved_circuits = tfq.layers.ResolveParameters(
    circuit_tensor, symbols_tensor, bitstring_tensor)
```
where `tfq.layers.ResolveParameters` would live next to `tfq.layers.AddCircuit` in the circuit construction module.  It would fill in any subset of symbols.

MichaelBroughton:
This is an interesting idea, do you think we need a full blown layer (i.e. are there multiple differen input signatures we want to support ? ) Also this layer would block differentiability on symbols that get resolved in them along the way (since there would be no way to pass them back). Are we ok with this ?

zaqqwerty:
I can imagine a few input signatures, output always 1-D tensor of circuits:

[...]

PR #322, opened 2020-07-27
Circuit parameter filler

zaqqwerty:
Adds the op called for in #286 for resolving parameters in-graph in circuits

### DIFF
--- a/tensorflow_quantum/core/ops/tfq_resolve_parameters_op.cc
+++ b/tensorflow_quantum/core/ops/tfq_resolve_parameters_op.cc
@@ -23,6 +23,7 @@ limitations under the License.
 #include "tensorflow/core/lib/core/status.h"
 #include "tensorflow/core/lib/core/threadpool.h"
 #include "tensorflow_quantum/core/ops/parse_context.h"
+#include "tensorflow_quantum/core/ops/tfq_simulate_utils.h"
 #include "tensorflow_quantum/core/src/program_resolution.h"
 
 namespace tfq {
--- a/tensorflow_quantum/core/ops/tfq_resolve_parameters_op.cc
+++ b/tensorflow_quantum/core/ops/tfq_resolve_parameters_op.cc
@@ -63,7 +63,7 @@ class TfqResolveParametersOp : public tensorflow::OpKernel {
       for (int i = start; i < end; i++) {
         Program program = programs[i];
         OP_REQUIRES_OK(context, ResolveSymbols(maps[i], &program, false));
-        programs.at(i).SerializeToString(&temp);
+        program.SerializeToString(&temp);
         output_tensor(i) = temp;
       }
     };
--- a/tensorflow_quantum/core/ops/tfq_resolve_parameters_op.cc
+++ b/tensorflow_quantum/core/ops/tfq_resolve_parameters_op.cc
@@ -15,11 +15,6 @@ limitations under the License.
 
 #include <string>
 
-#include "../qsim/lib/circuit.h"
-#include "../qsim/lib/gate_appl.h"
-#include "../qsim/lib/gates_cirq.h"
-#include "../qsim/lib/seqfor.h"
-#include "../qsim/lib/simmux.h"
 #include "cirq/google/api/v2/program.pb.h"
 #include "tensorflow/core/framework/op_kernel.h"
 #include "tensorflow/core/framework/shape_inference.h"
@@ -28,16 +23,12 @@ limitations under the License.
 #include "tensorflow/core/lib/core/status.h"
 #include "tensorflow/core/lib/core/threadpool.h"
 #include "tensorflow_quantum/core/ops/parse_context.h"
-#include "tensorflow_quantum/core/src/circuit_parser_qsim.h"
-#include "tensorflow_quantum/core/src/util_qsim.h"
+#include "tensorflow_quantum/core/src/program_resolution.h"
 
 namespace tfq {
 
 using ::cirq::google::api::v2::Program;
 using ::tensorflow::Status;
-
-typedef qsim::Cirq::GateCirq<float> QsimGate;
-typedef qsim::Circuit<QsimGate> QsimCircuit;
 
 class TfqResolveParametersOp : public tensorflow::OpKernel {
  public:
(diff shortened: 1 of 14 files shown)


zaqqwerty:
Hi Jae, I updated based on your comments, PTAL

tensorflow_quantum/core/ops/tfq_utility_ops_test.py
```diff
@@ -126,5 +128,234 @@ def test_padded_to_ragged(self, padded_array):
...
+                util.convert_to_tensor(circuit_list), symbol_names,
+                np.asarray(bitstring_list)))
+
+        expected_resolved_circuits = []
+        for circuit, resolver in zip(circuit_list, resolver_list):
+            expected_resolved_circuits.append(
+                cirq.resolve_parameters(circuit, resolver))
+
+        self.assertAllEqual(util.convert_to_tensor(expected_resolved_circuits),
+                            util.convert_to_tensor(test_resolved_circuits))
```
jaeyoo, line 282: Please compare them directly.

```python
self.assertAllEqual(expected_resolved_circuits, test_resolved_circuits)
```

Because protobuf doesn't guarantee that the serialized strings are encoded in the same order at every call.

zaqqwerty: Changed to comparing Cirq circuits directly

MichaelBroughton:
Can we also add a docstring somewhere for this op ?

zaqqwerty:
Hi @MichaelBroughton, I took the strategy of `tfq_unitary_op` to add the alias and docstring together, PTAL

tensorflow_quantum/core/ops/tfq_resolve_parameters_op.cc
```diff
@@ -0,0 +1,106 @@
...
+        output_tensor(i) = temp;
+      }
+    };
+
+    const int output_dim_size = programs.size();
+    const int block_size = GetBlockSize(context, output_dim_size);
+    context->device()
+        ->tensorflow_cpu_worker_threads()
+        ->workers->TransformRangeConcurrently(block_size, output_dim_size,
+                                              DoWork);
```
MichaelBroughton, line 76: We want to try and use `ParallelFor` here instead of `TransformRangeConcurrently`.

tensorflow_quantum/core/ops/tfq_utility_ops.py
```diff
@@ -58,3 +58,36 @@ def padded_to_ragged2d(masked_state):
     col_mask = tf.abs(tf.cast(masked_state[:, 0], tf.float32)) < 1.1
     masked = tf.ragged.boolean_mask(masked_state, col_mask)
     return padded_to_ragged(masked)
+
+
+def resolve_parameters(programs, symbol_names, symbol_values):
+    """Replace symbols in a batch of programs with concrete values.
+
```
MichaelBroughton, line 65: Can we also document the fact that this has the ability to "partially resolve" given parameters and therefore does not require you provide resolution values for ALL symbols in the circuit. I feel like a note that this op is NOT differentiable would also help our users understand that this op has very specific use cases.

tensorflow_quantum/core/ops/tfq_utility_ops_test.py
```diff
@@ -126,5 +128,235 @@ def test_padded_to_ragged(self, padded_array):
...
+                                               symbol_values_array)
+
+        # TODO(zaqqwerty): reapply test when ResolveSymbols defaults updated.
+        # with self.assertRaisesRegex(tf.errors.InvalidArgumentError,
+        #                             "Parameter map contains symbols not "
+        #                             "present in the program."):
+        #     # symbol_names tensor has the right type, but invalid value.
+        #     tfq_utility_ops.resolve_parameters(
+        #         util.convert_to_tensor(circuit_batch), ['junk'],
+        #         symbol_values_array)
```
MichaelBroughton, line 224: Block comment.

tensorflow_quantum/core/ops/tfq_utility_ops_test.py
```diff
@@ -126,5 +128,235 @@ def test_padded_to_ragged(self, padded_array):
...
+                    elif isinstance(tg, cirq.FSimGate):
+                        self._compare_gate_parameters(tg.theta, eg.theta)
+                        self._compare_gate_parameters(tg.phi, eg.phi)
+                    elif isinstance(
+                            tg, (cirq.PhasedXPowGate, cirq.PhasedISwapPowGate)):
+                        self._compare_gate_parameters(tg._global_shift,
+                                                      eg._global_shift)
+                        self._compare_gate_parameters(tg._exponent,
+                                                      eg._exponent)
+                        self._compare_gate_parameters(tg._phase_exponent,
```
MichaelBroughton, line 352: Can we not use gate equality checking with the `==` operation ?

tensorflow_quantum/core/src/program_resolution.h
```diff
@@ -42,12 +42,14 @@ tensorflow::Status ResolveQubitIds(
 // Resolves all of the symbols present in the Program. Iterates through all
 // operations in all moments, and if any Args have a symbol, replaces the one-of
 // with an ArgValue representing the value in the parameter map keyed by the
-// symbol. Returns an error if the parameter value cannot be found.
-// TODO(pmassey): Consider returning an error if a value in the parameter map
-// isn't used.
+// symbol. When `resolve_all` is true, returns an error if a symbol does not
+// have a correponding value in `param_map`.  When `flag_unused` is true,
+// returns an error if a value in the parameter map isn't used.
+// TODO(zaqqwerty): Work to flip default of `flag_unused` to true.
```
MichaelBroughton, line 48: In adding `flag_unused` we are making room for new error mechanisms that are not required by this PR and are not required currently by the rest of TFQ. What's the motivation for adding them here and now ?

tensorflow_quantum/core/ops/tfq_utility_ops_test.py
zaqqwerty: Gates are only approximately equal here (seems by difference of float and double); and, there was additionally difficulty getting expressions like `0.1*pi*a` and `0.314*a` to compare equal, which is taken care of in the sympy parsing at line 143

tensorflow_quantum/core/ops/tfq_utility_ops_test.py
zaqqwerty: Removed

tensorflow_quantum/core/ops/tfq_utility_ops_test.py
MichaelBroughton: maybe instead of having this intricate comparison function, we could just send both sides through the `convert_to_tensor` / `convert_from_tensor` pair until they equalize. In this case, something like:

```python3
expected_ans = tfq.from_tensor(tfq.to_tensor(resolved_circuits))
self.assertAllEqual(expected_ans, calculated_ans)
```

This way we add in all the same artifacts from going to and from tensor without having to have this equality checking code laying around.

tensorflow_quantum/core/ops/tfq_resolve_parameters_op.cc
zaqqwerty: Changed to `ParallelFor`

tensorflow_quantum/core/ops/tfq_utility_ops.py
zaqqwerty: Expanded docstring

tensorflow_quantum/core/ops/tfq_utility_ops_test.py
zaqqwerty: when I do this conversion equalizing, as
```python
tensor_test = util.from_tensor(util.convert_to_tensor(util.from_tensor(util.convert_to_tensor(test_resolved_circuits))))
tensor_expected = util.from_tensor(util.convert_to_tensor(util.from_tensor(util.convert_to_tensor(expected_resolved_circuits))))
self.assertAllEqual(tensor_test, tensor_expected)
```
For a batch and moment size of 2 in the circuit randomizer I get the error that the following terms are not equal:
```(cirq.XX**0.410076).on(cirq.GridQubit(0, 0), cirq.GridQubit(0, 1))```
```(cirq.XX**0.410075).on(cirq.GridQubit(0, 0), cirq.GridQubit(0, 1))```
Since the exponents are slightly different.  Using `self.assertAllClose` fails when encountering unresolved symbols.

Comment (zaqqwerty) [reply tensorflow_quantum/core/ops/tfq_utility_ops_test.py:352]: Trying rounding inside the util function now instead

Comment (MichaelBroughton) [reply tensorflow_quantum/core/ops/tfq_utility_ops_test.py:352]: Did `cirq.approx_eq` also not work ?

Comment (zaqqwerty) [reply tensorflow_quantum/core/ops/tfq_utility_ops_test.py]: adding rounding to 3 decimal places to the symbol values in  `util.random_symbol_circuit_resolver_batch` did not help resolve this either, added back the parsing for now

File: tensorflow_quantum/core/src/program_resolution.h (hunk 6 above)
Comment (zaqqwerty) [reply tensorflow_quantum/core/src/program_resolution.h]: Left over from some other attempts at things, reverted it now.

File: tensorflow_quantum/python/util.py [hunk 7]
```diff
@@ -63,9 +70,16 @@ def random_symbol_circuit(qubits,
...
+                used_symbols.add(this_random_symbol)
                 circuit[:i] += (
                     op**((np.random.random() if include_scalars else 1.0) *
-                         sympy.Symbol(np.random.choice(symbols)))).on(*locs)
+                         this_random_symbol)).on(*locs)
+
+    # Use the rest of the symbols
+    for s in all_symbols - used_symbols:
+        circuit += cirq.Circuit(cirq.H(qubits[0])**s)
+
```
MichaelBroughton, line 82: We can probably do this a bit cleaner with something like this:

```python3
random_symbols = list(symbols)
random.shuffle(random_symbols)
location = 0
```
Then in the else:
```python3
circuit[:i] += op ** ((np.random.random() if include_scalars else 1.0) * sympy.Symbol(random_symbols[location % len(random_symbols)]))
location += 1
```

The padding at the end is very similar to the original approach. This way you don't have to do set lookup or set difference at all.

tensorflow_quantum/core/src/program_resolution.cc
```diff
@@ -115,20 +115,24 @@ Status ResolveQubitIds(Program* program, unsigned int* num_qubits,
 
 Status ResolveSymbols(
     const absl::flat_hash_map<std::string, std::pair<int, float>>& param_map,
-    Program* program) {
+    Program* program, bool resolve_all /*=true*/) {
+  absl::flat_hash_set<std::string> used_symbols;
```
MichaelBroughton, line 119: Is this being used ?

tensorflow_quantum/core/src/program_resolution.cc
```diff
@@ -115,20 +115,24 @@ Status ResolveQubitIds(Program* program, unsigned int* num_qubits,
...
-                tensorflow::error::INVALID_ARGUMENT,
-                "Could not find symbol in parameter map: " + arg.symbol());
+            if (resolve_all) {
+              return Status(
+                  tensorflow::error::INVALID_ARGUMENT,
+                  "Could not find symbol in parameter map: " + arg.symbol());
+            }
+          } else {
+            arg.mutable_arg_value()->set_float_value(iter->second.second);
+            used_symbols.insert(arg.symbol());
```
MichaelBroughton, line 134: Can streamline this if with just:

```C++
if(iter == param_map.end()) {
  if(resolve_all){
    . . .
  }
  continue;
}
arg.mutable_arg_value()->set_float_value(iter->second.second)

```

tensorflow_quantum/core/src/program_resolution.cc
zaqqwerty: Good catch, removed

tensorflow_quantum/core/src/program_resolution.cc
zaqqwerty: Updated

tensorflow_quantum/python/util.py
zaqqwerty: Changed to remove the sets

zaqqwerty:
Using `cirq.approx_eq` also does not work.  I [opened an issue](https://github.com/quantumlib/Cirq/issues/3192) about this on cirq.

PR #323, opened 2020-07-31
build update

zaqqwerty:
removed some unused packages

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -116,14 +116,3 @@ cc_library(
     ],
 )
 
-http_archive(
-    name = "six_archive",
-    build_file = "@com_google_protobuf//:six.BUILD",
-    sha256 = "105f8d68616f8248e24bf0e9372ef04d3cc10104f1980f54d57b2ce73a5ad56a",
-    url = "https://pypi.python.org/packages/source/s/six/six-1.10.0.tar.gz#md5=34eed507548117b2ab523ab14b2f8b55",
-)
-
-bind(
-    name = "six",
-    actual = "@six_archive//:six",
-)
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -97,22 +97,3 @@ tf_workspace(tf_repo_name = "@org_tensorflow")
 load("//third_party/tf:tf_configure.bzl", "tf_configure")
 
 tf_configure(name = "local_config_tf")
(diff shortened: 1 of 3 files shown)


WORKSPACE
```diff
@@ -97,33 +97,3 @@ tf_workspace(tf_repo_name = "@org_tensorflow")
...
-  visibility = ["//visibility:public"],
-)
-    """,
-    sha256 = "7e7a57e33c59280a17a66e521396cd8b1a55d0676c9f807078522fda52114b5c",
-    strip_prefix = "eigen-eigen-8071cda5714d",
-    urls = [
-        "https://storage.googleapis.com/mirror.tensorflow.org/bitbucket.org/eigen/eigen/get/8071cda5714d.tar.gz",
-        "https://bitbucket.org/eigen/eigen/get/8071cda5714d.tar.gz",
-    ],
-)
```
jaeyoo, line 117: Wait, we didn't remove the `qsim_old` at all, so still there is a dependency on the `Eigen` in [here](https://github.com/tensorflow/quantum/blob/0c90033d8763478300a6626b038c7c7ddacb7a4a/tensorflow_quantum/core/qsim/unitary_space_test.cc#L18-L21). Shall we remove this BUILD lines after removing the entire dependent codes?

tensorflow_quantum/core/qsim/BUILD
```diff
@@ -174,7 +174,6 @@ cc_test(
         ":unitary_space",
         "//tensorflow_quantum/core/src:circuit",
         "@com_google_googletest//:gtest_main",
-        "@eigen//:eigen3",
```
jaeyoo, line 177: Is it really ok even if the `qsim_old` code requires `Eigen::**`?

zaqqwerty:
Good catch, seems we should not remove these until we remove `qsim_old`.  @MichaelBroughton are we good to remove the `qsim_old` yet?  I saw Unitary Feature got merged into external QSim, wasn't sure which pieces were still moving there

MichaelBroughton:
We’re getting close on external qsim. They need a few more tweaks and then to cut a release. From there we update our unitary op. Then we can cut qsim_old :).

zaqqwerty:
Will close pending `qsim_old` removal

PR #333, opened 2020-08-05
Added tf 2.1.0 under extras.

MichaelBroughton:
Adds the ability to do `python3 -m pip install tensorflow-quantum[extras]` which will also include a known-working version of `tf` version to depend on if the user doesn't have a nightly installed.

### DIFF
--- a/release/setup.py
+++ b/release/setup.py
@@ -51,6 +51,10 @@ class InstallPlatlib(install):
 
 
 REQUIRED_PACKAGES = ['cirq == 0.8.0', 'sympy == 1.5']
+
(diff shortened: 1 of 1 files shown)


release/setup.py
```diff
@@ -51,6 +51,10 @@ def finalize_options(self):
 
 
 REQUIRED_PACKAGES = ['cirq == 0.8.0', 'sympy == 1.5']
+
+# placed as extra to not have required overwrite existing nightly installs if
+# they exist.
+EXTRA_PACKAGES = ['tensorflow == 2.1.0']
```
yashk2810, line 57: Remove the space around `==`?

PR #339, opened 2020-08-12
New tutorial/example for qGAN

oliverob:
I have written a new tutorial/example jupyter notebook for implementing a qGAN using TFQ and TF-GAN. However, I have been stuck for a while trying to actually get it to work. My network doesn't seem to be learning correctly, and I am not sure how to fix it. If anyone could give me any pointers that would be brilliant. Currently, my discriminator loss is growing out of control, whilst my generator loss is shrinking. I am worried that the problem might actually be caused by a bug in the implementation that I can't find rather than something like the learning rates but I am not sure.


jaeyoo:
Thank you for this awesome contribution. First of all, I saw that `tensorflow_gan` is not installed. could you install? in notebook, we can get it by `!pip install -U tensorflow_gan`. Please let me review this notebook in a week, due to limited bandwidth. Thanks!

oliverob:
Thanks for all the comments. Sorry about all the lint and comment issues. I didn't mean for this to be production code yet - I was hoping to get try and figure out why it doesn't work before making the surrounding explaination correct and understandable. I'll make your suggested changes, and add a few of my own from work I did last week. I still cannot get it to learn correctly, I think I am approaching batches wrong. I'll let you know when I commit my newest version with my confused attempt at using batches.

docs/tutorials/qGAN.ipynb
```diff
@@ -0,0 +1,2644 @@
+{
```
oliverob, line 1: <p>I don't know if I am generating the correct format of data. I have tried a whole mix of options from averaging over a number of repetitions to the frequency of occurance of each number but I am not sure what is correct. This version seemed to make the most sense to me as the discriminator will see a sample which is a set of outputs of a circuit with the generator weights and so should be able to deduce the distribution from that. I thought that averaging might lose information about the distribution.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/quantum/pull/339/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

oliverob, line 1: <p>I need to change this to reflect the fact I removed the layer with 50 nodes. Though I'm not sure if I should add it back as removing it didn't seem to make a difference.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/quantum/pull/339/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

oliverob, line 1: <p>I need to add an explaination of how this all works. However, it currently doesn't work, so I cannot write it yet</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/quantum/pull/339/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

oliverob, line 1: <p>If it was learning correctly these graphs would show it tending towards a lognormal distribution. However, as you can see it isn't learning correctly. I have tried fiddling around with the learning rate etc. but I can't seem to fix it. Also, I don't think the entropy calculation is working correctly but it isn't necessary - I might just remove it.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/quantum/pull/339/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

oliverob:
@jaeyoo I have made the changes you suggested and tried to get it working but it still isn't learning correctly. Please could you have a look at the comments I have left in reviewnb and suggest anything I could try to make it work.

oliverob:
@jaeyoo Any ideas?

jaeyoo:
> @jaeyoo Any ideas?

Sorry, I am on the long vacations... please let me check it later.. I apologize for the belated messages.

PR #340, opened 2020-08-12
Updated nightly to tf 2.3.0 and bazel 3.1.0.

MichaelBroughton:
When a new relaease is cut, we will need to upgrade the version in all of the tutorials and docs too. Just want to confirm with @yashk2810 that we won't upgrade anything in the `docs/` folder (including tutorials) until we cut a new pip release, then we will upgrade all the 2.1.0 -> 2.3.0 in there.


yashk2810:
LGTM

PR #345, opened 2020-08-17
Refactor differentiators

### LINKED ISSUE #299: User-facing call to parameter shift differentiator

zaqqwerty:
Puts differentiator logic into its own function to expose intermediate circuits to the user.  Resolves #299


docs/tutorials/gradients.ipynb
```diff
@@ -503,19 +519,19 @@
...
-      "outputs": [],
       "source": [
         "test_value = 0.\n",
         "\n",
         "print('Finite difference:', my_grad(pauli_z, test_value))\n",
         "print('Sin formula:      ', -np.pi * np.sin(np.pi * test_value))"
-      ]
+      ],
+      "execution_count": null,
+      "outputs": []
```
MichaelBroughton, line 534: Once you get the notebook working, don't forget to restart it and clear the outputs.

tensorflow_quantum/python/differentiators/differentiator.py
```diff
@@ -179,6 +179,27 @@ def refresh(self):
             del self.expectation_op
         return self
 
+    @abc.abstractmethod
+    def get_intermediate_logic(self, programs, symbol_names, symbol_values,
+                               pauli_sums):
+        """Returns circuits and values which are used by both diff types.
```
MichaelBroughton, line 185: Going to need more description here if a user is going to understand how to implement this effectively.

tensorflow_quantum/python/differentiators/linear_combination.py
```diff
@@ -96,16 +96,14 @@ def __init__(self, weights, perturbations):
...
     @tf.function
-    def differentiate_analytic(self, programs, symbol_names, symbol_values,
-                               pauli_sums, forward_pass_vals, grad):
+    def get_intermediate_logic(self, programs, symbol_names, symbol_values,
+                               pauli_sums):
+        """Returns copies of the input programs for each perturbed symbol.
 
-        # these get used a lot
-        n_symbols = tf.gather(tf.shape(symbol_names), 0)
+        See base class for Args.
```
MichaelBroughton, line 103: What is being returned here ? The base class just says what the args are but intermediate logic for each differentiation algorithm could be drastically different and should be documented.

tensorflow_quantum/python/differentiators/parameter_shift.py
```diff
@@ -102,57 +157,20 @@ def differentiate_analytic(self, programs, symbol_names, symbol_values,
...
+        total_programs = n_param_gates * n_programs * n_shifts * n_symbols
         n_tile = n_shifts * n_param_gates * n_symbols
-        flat_perturbations = tf.concat([
-            tf.reshape(
-                tf.tile(tf.expand_dims(symbol_values, 0),
-                        tf.stack([n_tile, 1, 1])), [total_programs, n_symbols]),
-            tf.expand_dims(flat_shifts, axis=1)
-        ],
-                                       axis=1)
         flat_ops = tf.reshape(
```
MichaelBroughton, line 171: This isn't being returned by `get_intermediate_logic` function even though it is essential for the expectation calculation below, we shouldn't force a user to have to think harder than they need to on how to call into the op in their own way should they want to.

tensorflow_quantum/python/differentiators/differentiator.py
zaqqwerty: Expanded the docstring

tensorflow_quantum/python/differentiators/linear_combination.py
zaqqwerty: Changed the API of `get_intermediate_logic`, the base class now describes both `Args` and `Returns`

tensorflow_quantum/python/differentiators/parameter_shift.py
zaqqwerty: Refactored

zaqqwerty:
@MichaelBroughton I came up with a new API for the `get_intermediate_logic` function which I think provides the flexibility needed to implement SG diffs cleanly.  I implemented it fully for `LinearCombination`, but I wanted to get your thoughts on it before I go ahead with the unit tests and implementation for PS.   Also, I forked off the initial removal of SGD stuff into its own PR #372 so that this one doesn't get too cluttered.

tensorflow_quantum/python/differentiators/differentiator.py
```diff
@@ -179,6 +179,73 @@ def refresh(self):
...
+                The first dimension is the length of the input `programs`, the
+                second dimension is the length of the second dimension of the
+                input `pauli_sum`, the third dimension is the length of the
+                input `symbol_names`, and the last two dimensions are the same
+                shape as the output `flat_pauli_sums`.  The value at index
+                `ijkmn` is the amount of weight to give the measurement result
+                at `mn` when summing over expectation values to obtain the
+                gradient of the symbol `k` with respect to pauli sum `j` of
+                program `i`.
+        """
```
MichaelBroughton, line 247: Given the potentially drastically different "intermediate logic blocks" between differentiators. I think it might make more sense to just have the return signatures defined on a per differentiator basis and leave the parent class as an abstract method. Trying to fit all differentiators under this roof might prove pretty hard.

MichaelBroughton (changes requested):
Looking good, a major change we should adopt here is letting the differentiator define it's own return signature, as we introduce more it will be harder and harder to keep a standardized return signature. An advanced user would also likely know that and expect different forms as well.

tensorflow_quantum/python/differentiators/differentiator.py
zaqqwerty: Even though every differentiator can be defined via this interface, I understand your point that it could be a bit hard to reason about.  I think it could make sense to restrict the interface slightly to assume we will use the same gradient procedure for every circuit; then the mapper tensor would be the same for every circuit in the input batch, allowing us to drop an index and reduce the size of the mapper.  I'll think a bit more about how to simplify this interface along that line.

zaqqwerty: Changed to a new, simpler batch interface.  ParameterShift and LinearCombination have been updated to obey the interface.  I still need to write tests for `get_intermediate_logic` confirming shapes and values explicitly.

docs/tutorials/gradients.ipynb
zaqqwerty: This is still how it looks after going to "Runtime > Restart runtime" and then doing "Edit > Clear all outputs" before downloading

tensorflow_quantum/python/differentiators/differentiator.py
zaqqwerty: Now added an explicit equation at the end of the docstring to justify why the returns have the shapes specified.

zaqqwerty:
Given your comment on the interface difficulty, I simplified things a bit to a batch API, where the same transformation is applied to each circuit in the input.  I refactored `LinearCombination` and `ParameterShift` into compliance with this new interface.  I added an explicit equation at the end of the base class docstring to help the user understand the shapes of the returns and how to use them.  Finally, I added annotated tests for these new `get_intermediate_logic` functions.

MichaelBroughton (changes requested):
The interface change seems reasonable although I personally still think it is still pretty complicated.

A more major problem that I have with this re-implementation is that we have moved from making just ONE call to the provided expectation op to many seperate calls to the given expectation function inside of a `map_fn`. Looking at the docs page for `map_fn` we see: 

> map_fn unstacks elems on axis 0 to obtain a sequence of elements; calls fn to transform each element; and then stacks the transformed values back together.

In our case, this will drastically degrade performance both in simulation and on a real QPU. Simulation with have to allocate / simulate and deallocate memory for each individual circuit which is very wasteful. If the circuits are small it will also be unable to parallelize very much ( if at all ). When running on QPU the performance will also be worse. If we are unable to see the ENTIRE batch of circuits we need torun , we will never be able totake full advantage of more efficient engine methods like `run_batch` (from Cirq) or `run_sweep` (from Cirq).

If we want to keep this structure I think we should definitely fix this issue inside of the `differentiate_***` methods too.

zaqqwerty:
Sure, I can just stack the circuits first instead

zaqqwerty:
Reshaped things in the default diff logic so that expectation op is only called once per differentiation again

zaqqwerty:
Had to double the number of samples in the new gradient tests to prevent flaking

MichaelBroughton (commented):
I'm sorry that this has turned into such a long discussion. Perhaps it is a good indicator that we might need a design document....

Now I'm wondering how get_intermediate_logic might behave with something like the SGDifferentiator where the number of samples taken matters a great deal. The `get_intermediate_logic` method does not take num_samples as an argument which might be something we want to include ?

Would you be open to discussing this more in tomorrow meeting ?

zaqqwerty:
Sure, let's chat about it tomorrow, I want to make sure it works for everyone's use cases.  I spoke with Jae about this new interface, he was of the opinion it will ease the re-implementation of the SGDifferentiators (those also used just a simple tiling of the input num_samples, but we can discuss including it as an argument to `get_intermediate_logic`)

zaqqwerty:
I thought more about your comment on `num_samples`.  I was able to include it as part of the input and output of `get_intermediate_logic`, and this actually allowed the definitions of `differentiate_analytic` and `differentiate_sampled` to be pulled all the way to the base class only, greatly reducing code duplication.

MichaelBroughton (changes requested):
Looking at the code I remember how tricky it was implementing these things the first time. Would we be open to doing this changeover in stages:

1. Add the interface and functionality of `get_intermediate_logic`. (This PR)
2. Move over all of the `differentiate_analytic` and `differentiate_sampled` to use them (another PR) ?

tensorflow_quantum/python/differentiators/differentiator.py
```diff
@@ -180,14 +180,111 @@ def refresh(self):
...
+
+            As an explicit equation, the return values are defined in terms of
+            the argument values such that:
+
+            d(<programs[i](symbol_values[i])|
+                pauli_sums[i, j]|programs[i](symbol_values[i])>)
+            / d(symbol_names[k]) = sum_mn (batch_mapper[i, j, k, m, n] *
+                <batch_programs[i][m](batch_symbol_values[i][m])|
+                    batch_pauli_sums[i, m, n]
+                        |batch_programs[i][m](batch_symbol_values[i][m])>)
```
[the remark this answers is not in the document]
zaqqwerty: Here `d` is the differentiation symbol

tensorflow_quantum/python/differentiators/differentiator.py
```diff
@@ -180,14 +180,111 @@ def refresh(self):
         return self
 
     @abc.abstractmethod
+    def get_intermediate_logic(self, programs, symbol_names, symbol_values,
+                               pauli_sums, num_samples):
+        """Returns components that can be used to build the gradient.
```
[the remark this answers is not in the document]
zaqqwerty: Slightly too long for first line only, added shortened version

tensorflow_quantum/python/differentiators/differentiator.py
```diff
@@ -180,14 +180,111 @@ def refresh(self):
...
+        the input batch will have the same gradient procedure applied to it.
+        Thus the outer dimension of all returned components is the same as the
+        input `programs` batch dimension.
+
+        For each entry in the batch dimension of the returned values,
+        inheriting differentiators are free to generate new circuits, symbols,
+        and measurement ops as needed. To combine these outputs into the
+        required derivatives, the last return value is a map specifying the
+        linear combination of batched expectation values to use to calculate the
+        derivative of each input parameter.
```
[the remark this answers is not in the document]
zaqqwerty: Added a `NOTE:` paragraph pointing this out

tensorflow_quantum/python/differentiators/differentiator.py
```diff
@@ -180,14 +180,111 @@ def refresh(self):
...
+                `symbol_values` should be placed inside of the circuits in
+                `programs`.
+            symbol_values: `tf.Tensor` of real numbers with shape
+                [batch_size, n_params] specifying parameter values to resolve
+                into the circuits specified by programs, following the ordering
+                dictated by `symbol_names`.
+            pauli_sums: `tf.Tensor` of strings with shape [batch_size, n_ops]
+                containing the string representation of the operators that will
+                be used on all of the circuits in the expectation calculations.
+            num_samples: `tf.Tensor` of positive integers representing the
```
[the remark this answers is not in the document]
zaqqwerty: Added the shape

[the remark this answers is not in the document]
zaqqwerty: Makes sense to make this optional; do you know of any best practices around optional arguments to functions with `tf.function`?  My experience is these can get finicky

zaqqwerty:
Hi @MichaelBroughton, I addressed your comments on the docstring.  I also broke this PR in two as you suggested; this one just adds the `get_intermediate_logic` interface, implementations, and tests; #379 deals with removing the old `differentiate_analytic` and `differentiate_sampled` code.

tensorflow_quantum/python/differentiators/differentiator.py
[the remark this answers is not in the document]
MichaelBroughton: I'm pretty sure tf.functions don't support the intuitive python style default arguments, but an easy workaround is to just have a null value tensor that indicates the `get_intermeidate_logic` function won't process the samples tensor. For example you could pass in tf.constant(-2) with shape = `()` to indicate you don't need samples.

tensorflow_quantum/python/differentiators/differentiator.py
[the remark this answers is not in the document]
MichaelBroughton: Might want to clear that up so it's easy to understand in the doc renderings.

MichaelBroughton (changes requested):
Tested this locally on my machine and made a worrying discovery: In a relatively moderate setting of a batch of 64 circuits each with 25 symbols , 10 ops and a central-difference differentiatior, just calling `get_intermediate_logic` consumed a huge amount of memory. I worked out the numbers and I think the `batch_mapper` is the culprit with shape [64, 10, 25, 50, 50] =>  40 million entries (the last two might be 25's ?) which is far too many. 

I think we need to reconsider this component. Is there any way we could shrink down this tensor and maintain fully intuitive generality ? Maybe this makes a case for moving to each differentiator having it's own unique return signature on `get_intermediate_logic` ?

zaqqwerty:
@MichaelBroughton I think you're right that the discussion would be aided by a design doc.  I've started writing one (sent you a DM)

zaqqwerty:
After some more thought, since this PR just adds the functionality of intermediate logic and doesn't remove the way it functions elsewhere, could we put this in as-is (after the docstring and optional `num_samples` arg updates), then open an issue and do the careful design doc before swapping this logic into `differentiate_analytic` and `differentiate_sampled`?

To follow up on the memory issue, in your example, `batch_mapper` would actually have shape [64, 10, 25, 51, 10] which implies 8,160,000 entries.  This is about a factor of 100 larger than the existing intermediate steps of LinearCombination [here](https://github.com/tensorflow/quantum/blob/0712199eed1ce77a3b736d886c757ea48d41987f/tensorflow_quantum/python/differentiators/linear_combination.py#L152), `all_perturbations`, which for your example would have shape `[n_symbols, n_perturbations, n_programs, n_symbols]` ==> [25, 2, 64, 25] which is 80,000 entries.  If this is too large, one way that comes to mind to reduce the memory footprint might be to switch batch mapper to a [sparse tensors](https://www.tensorflow.org/api_docs/python/tf/sparse/SparseTensor), since most of the entries in the batch mapper (and in `all_perturbations`) in this case are zero.

MichaelBroughton:
I don't think taking on a factor 100x memory growth and then moving to make `differentiate_analytic` and `differentiate_sampled` depend on them would be a wise move here. Let's take a look at the design document and then circle back once we've found some solutions that work on all fronts.

zaqqwerty:
Completed design doc has a slightly different proposed interface, closing this one

PR #351, opened 2020-08-19
331 sample repetitions

### LINKED ISSUE #331: Specify repetitions circuitwise in Sample op
When using the `Sample` layer or op, there are sometimes workflows where one desires different numbers of samples from each circuit in the batch.  Currently, I get around this by applying the `Sample` layer once per circuit, each called with the correct number of repetitions.  It would be easier and likely be more performant to allow `repetitions` to be a tensor of same shape as the input circuits, where each entry specifies the number of samples to draw from the corresponding circuit.

zaqqwerty:
Allows number of samples to be specified independently for each circuit in the batch.  Resolves #331

### DIFF
--- a/tensorflow_quantum/python/layers/circuit_executors/sample.py
+++ b/tensorflow_quantum/python/layers/circuit_executors/sample.py
@@ -165,28 +165,34 @@ class Sample(tf.keras.layers.Layer):
             `inputs`, `symbol_names`, `symbol_values`:
                 see `input_checks.expand_circuits`
             `repetitions`: a Python `int` or a pre-converted
-                `tf.Tensor` containing a single `int` entry.
+                `tf.Tensor` containing a single `int` entry or is of shape
+                [batch_size].
 
         Output shape:
             `tf.RaggedTensor` with shape:
-                [batch size of symbol_values, repetitions, <ragged string size>]
+                [batch size of symbol_values, <ragged repetitions>, <ragged string size>]
                     or
-                [number of circuits, repetitions, <ragged string size>]
+                [number of circuits, <ragged repetitions>, <ragged string size>]
         """
         if repetitions is None:
             raise ValueError("Number of repetitions not specified.")
+
+        inputs, symbol_names, symbol_values = input_checks.expand_circuits(
+            inputs, symbol_names, symbol_values)
 
         # Ingest and promote repetitions.
         if isinstance(repetitions, numbers.Integral):
             if not repetitions > 0:
                 raise ValueError("Repetitions must be greater than zero.")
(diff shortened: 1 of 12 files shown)


tensorflow_quantum/python/layers/circuit_executors/sample.py
```diff
@@ -165,28 +165,35 @@ def call(self,
...
+        inputs, symbol_names, symbol_values = input_checks.expand_circuits(
+            inputs, symbol_names, symbol_values)
+
         # Ingest and promote repetitions.
         if isinstance(repetitions, numbers.Integral):
             if not repetitions > 0:
                 raise ValueError("Repetitions must be greater than zero.")
-            repetitions = tf.convert_to_tensor([repetitions], dtype=tf.int32)
+            repetitions = tf.fill(
+                tf.shape(inputs)[0], tf.constant(repetitions, dtype=tf.int32))
```
MichaelBroughton, line 188: The `tf.shape(inputs)[0]` will not work if this is compiled/crawled. You might want to switch to something simpler:

```python3
repetitions = tf.tile(tf.constant([repetitions], dtype=tf.int32), tf.shape(inputs))
```

tensorflow_quantum/python/layers/circuit_executors/sample.py
```diff
@@ -165,28 +165,35 @@ def call(self,
...
         if not tf.is_tensor(repetitions):
             raise TypeError("repetitions cannot be parsed to int32 tensor"
                             " tensor given input: ".format(repetitions))
 
-        inputs, symbol_names, symbol_values = input_checks.expand_circuits(
-            inputs, symbol_names, symbol_values)
+        if not all(tf.shape(repetitions) == tf.shape(inputs)):
+            raise ValueError("repetitions must be the same shape as inputs")
+
+        repetitions = tf.cast(repetitions, tf.int32)
```
MichaelBroughton, line 197: These shape check should happen down in the ops. Right now it is catching errors that should propogate up  from op level. It will also not work if this statement is compiled/crawled.

MichaelBroughton (changes requested):
Looking good, I would say before we shoot to merge this PR, it might make sense to merge a change to `batch_util.py` that enables a variable number of samples to be sent back.

Outside of that there are a few missing behaviors and tests in the ops that should be pushed down to be consistent with our other ops. Some more careful testing on this feature expansion is also probably a good idea (ones like making sure that we are padding in the right places and the shape is what we expect it to be).

There's also this huge change to handling the quantum engine sampler which seems like it should be a seperate PR as well. Overall, we might want to tackle this and the changes in `batch_util.py` in seperate PRs first.

zaqqwerty:
Closing for now, I am doing this manually outside of library TFQ and it works well enough.

PR #356, opened 2020-08-20
Adjoint differentiatior part 1/n.

MichaelBroughton:
Adds basic qsim utilities needed by adjoint differentiation. In this PR I've added:

1. `ApplyGateDagger` like `ApplyGate` in qsim, but daggers the gate before paplying.
2. `ApplyFusedGateDagger` same as above.
3. `AccumulateOperators`. Applies a large operator matrix multiplication in place to a state. Takes you from `|psi> -> O|psi>` for any O

### DIFF
--- a/tensorflow_quantum/core/src/util_qsim_test.cc
+++ b/tensorflow_quantum/core/src/util_qsim_test.cc
@@ -333,5 +333,194 @@ TEST(UtilQsimTest, CompoundCase) {
   EXPECT_NEAR(exp_v, 4.1234, 1e-5);
 }
 
+TEST(UtilQsimTest, ApplyGateDagger) {
+  // Create circuit to prepare initial state.
+  QsimCircuit simple_circuit;
+  simple_circuit.num_qubits = 2;
+  simple_circuit.gates.push_back(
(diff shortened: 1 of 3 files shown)


tensorflow_quantum/core/src/util_qsim_test.cc
```diff
@@ -333,5 +333,194 @@ TEST(UtilQsimTest, CompoundCase) {
...
+  // 4.0 I
+  p_term_scratch = p_sum.add_terms();
+  p_term_scratch->set_coefficient_real(4.0);
+
+  // A second sum.
+  PauliSum p_sum2;
+  PauliTerm* p_term_scratch2 = p_sum2.add_terms();
+  p_term_scratch2->set_coefficient_real(-5.0);
+
+  AccumulateOperators({p_sum, p_sum2}, {0.5, 0.25}, sim, ss, sv, scratch);
```
jaeyoo: Sorry, I didn't get the result {0.5, 0.25} from 4.0 I and -5.0 I. Could you add some comments here?

jaeyoo (changes requested):
LGTM with minor questions.

Issue #256, opened 2020-06-06
Adjoint Differentiation


MichaelBroughton:
While not really viable for NiSQ circuits (outside of simulation), the request has been made a few times to implement the adjoint gradient calculation method for differentiation. This would roughly entail:

1. Writing a new C++ op to calculate gradients via the adjoint method
2. Writing a new differentiator that uses this op inside of it and only defines this operation for `differentiate_analytic`.

Going to leave this open for discussion for anyone who's interested.

refraction-ray:
This will definitely be a great feature especially in the context of variational circuit simulations where gradients can be obtained more quickly compared to finite difference or parameter shift (which is also somehow finite difference) when the number of variational params is large.

I also want to mention that backward AD in circuit case (reversible) can be implemented in a more memory efficient way, i.e. the memory required can be reduced to constant  (width of circuit) compared to general adjoint methods where the memory consumption is proportional to the depth of network/graph/circuit. One can recompute the intermediate value in backward pass since the computation is reversible without keep them in forward pass. This can in principle enable AD of very deep circuit.

MichaelBroughton:
> I also want to mention that backward AD in circuit case (reversible) can be implemented in a more memory efficient way, i.e. the memory required can be reduced to constant (width of circuit) compared to general adjoint methods where the memory consumption is proportional to the depth of network/graph/circuit. One can recompute the intermediate value in backward pass since the computation is reversible without keep them in forward pass. This can in principle enable AD of very deep circuit.

Would you mind elaborating on this method you are describing ? Suppose I have this circuit
```
(0, 0): ───H^alpha───@───
                     │
(0, 1): ───X^beta────X───
```
and I want to compute the gradient of `<psi | X(0,0) * X(0, 1) | psi>` w.r.t `alpha` and `beta` where lets say `alpha = 0.123` and `beta = 0.456`. Can you help me understand what you have in mind to save memory and make things more efficient here ( I didn't exactly follow what you mentioned above ) ?

refraction-ray:
Hi, @MichaelBroughton , I have just written some short notes on this but still in an abstract fashion. I don't know whether it is clear enough or make sense to you.  If you still have any questions, we can further discuss on that and maybe go through the detailed example you presented above.

<img width="903" alt="Screen Shot 2020-07-09 at 15 42 43" src="https://user-images.githubusercontent.com/35157286/87011661-cd62a400-c1fa-11ea-8f9f-b67287f6f7d2.png">
[...]

PR #360, opened 2020-08-24
Adjoint 3/n: Adding parsing functionality for adjoint needs.

MichaelBroughton:
Added the `GradientOfGate` struct and corresponding circuit analysis logic to be used by the adjoint method. This marks all the code that needs adding in the `src` directory. From here out it should be writing the C++ op code as well as the rest of the python implementation. Progressing on #256 .


jaeyoo (approved):
LGTM

PR #362, opened 2020-08-24
Adjoint 4/n: Added basic functioning op.

MichaelBroughton:
Adds a basic functioning op for the adjoint gradient. I'm thinking that we should NOT expose this op to the user and instead force them to make use of the differentiatiors interface with an appropriately linked op, that is why I haven't made any modifications to the `__init__.py` files. My thinking behind this is that if a user wanted to call into this op they could just as easily do an appropriate `with tf.GradientTape() ...` style statement that sticks more closely to the expected TF logic for getting out gradients instead of `tfq.adj_grad(...)`.

What do @zaqqwerty and @jaeyoo think ?

Yet more progress on #256 . One or two more PRs should do it.

### DIFF
--- a/tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
+++ b/tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
@@ -56,10 +56,10 @@ class TfqAdjointGradientOp : public tensorflow::OpKernel {
 
     // Create the output Tensor.
     const int output_dim_batch_size = context->input(0).dim_size(0);
-    const int output_dim_op_size = context->input(2).dim_size(1);
+    const int output_dim_param_size = context->input(2).dim_size(1);
     tensorflow::TensorShape output_shape;
     output_shape.AddDim(output_dim_batch_size);
-    output_shape.AddDim(output_dim_op_size);
+    output_shape.AddDim(output_dim_param_size);
 
     tensorflow::Tensor* output = nullptr;
     OP_REQUIRES_OK(context, context->allocate_output(0, output_shape, &output));
@@ -215,12 +215,14 @@ class TfqAdjointGradientOp : public tensorflow::OpKernel {
           if (j == partial_fused_circuits[i].size() - 1) {
             break;
           }
+          // Apply the original gate of the gradient_gate,
+          // not the gradient_gate itself.
           qsim::ApplyGate(
               sim, qsim_circuits[i].gates[gradient_gates[i][j].index], sv);
         }
 
         // sv now contains psi
-        // scratch contains (sum_i paulis_sums[i] * downstream_grads[i])|psi>
+        // scratch contains (sum_j paulis_sums[i][j] * downstream_grads[j])|psi>
         // scratch2 now contains psi as well.
         AccumulateOperators(pauli_sums[i], downstream_grads[i], sim, ss, sv,
                             scratch2, scratch);
@@ -247,6 +249,11 @@ class TfqAdjointGradientOp : public tensorflow::OpKernel {
             // don't need not-found check since this is done upstream already.
             const auto it = maps[i].find(gradient_gates[i][j - 1].params[k]);
             const int loc = it->second.first;
+            // Apply finite differencing for adjoint gradients.
+            // Finite differencing enables applying multiple `gradient_gate`
+            // of a symbol at the same circuit. For analytic methods like
+            // parameter-shift we need to apply a single `gradient_gate`
+            // per a symbol.
             (*output_tensor)(i, loc) += ss.RealInnerProduct(sv, scratch) +
                                         ss.RealInnerProduct(scratch, sv);
 
@@ -316,12 +323,14 @@ class TfqAdjointGradientOp : public tensorflow::OpKernel {
         if (j == partial_fused_circuits[i].size() - 1) {
           break;
         }
+        // Apply the original gate of the gradient_gate,
+        // not the gradient_gate itself.
         qsim::ApplyGate(sim, qsim_circuits[i].gates[gradient_gates[i][j].index],
                         sv);
       }
 
       // sv now contains psi
-      // scratch contains (sum_i paulis_sums[i] * downstream_grads[i])|psi>
+      // scratch contains (sum_j paulis_sums[i][j] * downstream_grads[j])|psi>
       // scratch2 now contains psi as well.
       AccumulateOperators(pauli_sums[i], downstream_grads[i], sim, ss, sv,
                           scratch2, scratch);
@@ -348,6 +357,11 @@ class TfqAdjointGradientOp : public tensorflow::OpKernel {
           // don't need not-found check since this is done upstream already.
           const auto it = maps[i].find(gradient_gates[i][j - 1].params[k]);
           const int loc = it->second.first;
+          // Apply finite differencing for adjoint gradients.
+          // Finite differencing enables applying multiple `gradient_gate`
+          // of a symbol at the same circuit. For analytic methods like
+          // parameter-shift we need to apply a single `gradient_gate`
+          // per a symbol.
           (*output_tensor)(i, loc) += ss.RealInnerProduct(sv, scratch) +
                                       ss.RealInnerProduct(scratch, sv);
 
--- a/tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
+++ b/tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
@@ -1,0 +1,403 @@
+/* Copyright 2020 The TensorFlow Quantum Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include <memory>
+#include <vector>
+
+#include "../qsim/lib/circuit.h"
+#include "../qsim/lib/gate_appl.h"
+#include "../qsim/lib/gates_cirq.h"
+#include "../qsim/lib/seqfor.h"
+#include "../qsim/lib/simmux.h"
+#include "cirq/google/api/v2/program.pb.h"
+#include "tensorflow/core/framework/op_kernel.h"
+#include "tensorflow/core/framework/shape_inference.h"
+#include "tensorflow/core/framework/tensor_shape.h"
+#include "tensorflow/core/lib/core/error_codes.pb.h"
+#include "tensorflow/core/lib/core/status.h"
+#include "tensorflow/core/lib/core/threadpool.h"
+#include "tensorflow_quantum/core/ops/parse_context.h"
+#include "tensorflow_quantum/core/proto/pauli_sum.pb.h"
+#include "tensorflow_quantum/core/src/adj_util.h"
+#include "tensorflow_quantum/core/src/util_qsim.h"
+
+namespace tfq {
+
+using ::cirq::google::api::v2::Program;
+using ::tensorflow::Status;
+using ::tfq::proto::PauliSum;
+
+typedef qsim::Cirq::GateCirq<float> QsimGate;
+typedef qsim::Circuit<QsimGate> QsimCircuit;
+
+class TfqAdjointGradientOp : public tensorflow::OpKernel {
+ public:
+  explicit TfqAdjointGradientOp(tensorflow::OpKernelConstruction* context)
+      : OpKernel(context) {}
+
+  void Compute(tensorflow::OpKernelContext* context) override {
+    // TODO (mbbrough): add more dimension checks for other inputs here.
+    const int num_inputs = context->num_inputs();
+    OP_REQUIRES(context, num_inputs == 5,
+                tensorflow::errors::InvalidArgument(absl::StrCat(
+                    "Expected 5 inputs, got ", num_inputs, " inputs.")));
+
+    // Create the output Tensor.
+    const int output_dim_batch_size = context->input(0).dim_size(0);
+    const int output_dim_op_size = context->input(2).dim_size(1);
+    tensorflow::TensorShape output_shape;
+    output_shape.AddDim(output_dim_batch_size);
+    output_shape.AddDim(output_dim_op_size);
+
+    tensorflow::Tensor* output = nullptr;
+    OP_REQUIRES_OK(context, context->allocate_output(0, output_shape, &output));
+    auto output_tensor = output->matrix<float>();
+
+    // Parse program protos.
+    std::vector<Program> programs;
+    std::vector<int> num_qubits;
+    std::vector<std::vector<PauliSum>> pauli_sums;
+    OP_REQUIRES_OK(context, GetProgramsAndNumQubits(context, &programs,
+                                                    &num_qubits, &pauli_sums));
+
+    std::vector<SymbolMap> maps;
+    OP_REQUIRES_OK(context, GetSymbolMaps(context, &maps));
+
+    OP_REQUIRES(context, pauli_sums.size() == programs.size(),
+                tensorflow::errors::InvalidArgument(absl::StrCat(
+                    "Number of circuits and PauliSums do not match. Got ",
+                    programs.size(), " circuits and ", pauli_sums.size(),
+                    " paulisums.")));
+
+    // Construct qsim circuits.
+    std::vector<QsimCircuit> qsim_circuits(programs.size(), QsimCircuit());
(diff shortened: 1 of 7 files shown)


tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
```diff
@@ -0,0 +1,403 @@
...
+            ss.CopyState(sv, scratch2);
+            qsim::ApplyGate(sim, gradient_gates[i][j - 1].grad_gates[k], sv);
+
+            // don't need not-found check since this is done upstream already.
+            const auto it = maps[i].find(gradient_gates[i][j - 1].params[k]);
+            const int loc = it->second.first;
+            (*output_tensor)(i, loc) += ss.RealInnerProduct(sv, scratch) +
+                                        ss.RealInnerProduct(scratch, sv);
+
+            ss.CopyState(scratch2, sv);
```
jaeyoo, line 253: it seems like applying the `gradient_gates[i][j-1].grad_gates[k]` to the statevector, and calculate the innerproduct and then rollback the statevector before applying the `grad_gates[k]`. It looks like `scratch2` is always a temporal backup of `sv`. If it is the case, shall we reduce the number of `CopyState()` from `2` to `1`? I mean:
```cc
ss.CopyState(sv, scratch2);
qsim::ApplyGate(sim, gradient_gates[i][j - 1].grad_gates[k], scratch2); /* it was `sv`*/

// don't need not-found check since this is done upstream already.
const auto it = maps[i].find(gradient_gates[i][j - 1].params[k]);
const int loc = it->second.first;
(*output_tensor)(i, loc) += ss.RealInnerProduct(scratch2, scratch) +
                            ss.RealInnerProduct(scratch, scratch2);   /* <scratch2 | scratch> + <scratch|scratch2> */
```

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
```diff
@@ -0,0 +1,403 @@
...
+            // Copy sv onto scratch2 in anticipation of non-unitary "gradient
+            // gate".
+            ss.CopyState(sv, scratch2);
+            qsim::ApplyGate(sim, gradient_gates[i][j - 1].grad_gates[k], sv);
+
+            // don't need not-found check since this is done upstream already.
+            const auto it = maps[i].find(gradient_gates[i][j - 1].params[k]);
+            const int loc = it->second.first;
+            (*output_tensor)(i, loc) += ss.RealInnerProduct(sv, scratch) +
+                                        ss.RealInnerProduct(scratch, sv);
```
jaeyoo, line 251: I think we need some comments like the following for the future (maybe) analytic gradient methods:
```
// Apply finite differencing for adjoint gradients.
// Finite differencing enables applying multiple `gradient_gate` of a symbol at the same circuit.
// For analytic methods like parameter-shift we need to apply a single `gradient_gate`
// per a symbol.
```

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
```diff
@@ -0,0 +1,403 @@
...
+        ss.SetStateZero(sv);
+        for (int j = 0; j < partial_fused_circuits[i].size(); j++) {
+          for (int k = 0; k < partial_fused_circuits[i][j].size(); k++) {
+            qsim::ApplyFusedGate(sim, partial_fused_circuits[i][j][k], sv);
+          }
+          if (j == partial_fused_circuits[i].size() - 1) {
+            break;
+          }
+          qsim::ApplyGate(
+              sim, qsim_circuits[i].gates[gradient_gates[i][j].index], sv);
```
jaeyoo, line 219: Could you add a comment? I confused at the first glance of this code.
```
// Apply the original gate of the gradient_gate, not the gradient_gate itself.
```

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
```diff
@@ -0,0 +1,403 @@
...
+      ss.SetStateZero(sv);
+      for (int j = 0; j < partial_fused_circuits[i].size(); j++) {
+        for (int k = 0; k < partial_fused_circuits[i][j].size(); k++) {
+          qsim::ApplyFusedGate(sim, partial_fused_circuits[i][j][k], sv);
+        }
+        if (j == partial_fused_circuits[i].size() - 1) {
+          break;
+        }
+        qsim::ApplyGate(sim, qsim_circuits[i].gates[gradient_gates[i][j].index],
+                        sv);
```
jaeyoo, line 320: Please add a comment for clarification
```
// Apply the original gate of the gradient_gate, not the gradient_gate itself.
```

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
```diff
@@ -0,0 +1,403 @@
...
+      }
+
+      // sv now contains psi
+      // scratch contains (sum_i paulis_sums[i] * downstream_grads[i])|psi>
```
jaeyoo, line 324: `sum_j pauli_sums[i][j] * downstream_grads[i] |psi>`?

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
```diff
@@ -0,0 +1,403 @@
...
+          ss.CopyState(sv, scratch2);
+          qsim::ApplyGate(sim, gradient_gates[i][j - 1].grad_gates[k], sv);
+
+          // don't need not-found check since this is done upstream already.
+          const auto it = maps[i].find(gradient_gates[i][j - 1].params[k]);
+          const int loc = it->second.first;
+          (*output_tensor)(i, loc) += ss.RealInnerProduct(sv, scratch) +
+                                      ss.RealInnerProduct(scratch, sv);
+
+          ss.CopyState(scratch2, sv);
```
jaeyoo, line 354: can we remove one `CopyState()` as mentioned above?

tensorflow_quantum/core/ops/tfq_adj_grad_op_test.py
```diff
@@ -0,0 +1,338 @@
...
+            util.convert_to_tensor(circuit_batch),
+            tf.convert_to_tensor(symbol_names),
+            tf.convert_to_tensor(symbol_values_array),
+            util.convert_to_tensor(op_batch), prev_grads)
+
+        self.assertAllClose(out,
+                            np.array([[-2.100, -1.7412, -1.5120]]),
+                            atol=1e-3)
+
+    def test_calculate_adj_grad_simple_case3(self):
```
jaeyoo, line 300: I saw that there is no testing case like having one symbol with multiple parameterized gates. Could you please add the case?
```python
        n_qubits = 2
        batch_size = 1
        symbol_names = ['alpha']
        qubits = cirq.GridQubit.rect(1, n_qubits)
circuit_batch, resolver_batch = \
        [cirq.Circuit(cirq.X(qubits[0]) ** sympy.Symbol('alpha'),
            cirq.Y(qubits[1]) ** sympy.Symbol('alpha'),
            cirq.CNOT(qubits[0], qubits[1]),
            cirq.FSimGate(
                sympy.Symbol('alpha'),
                sympy.Symbol('alpha'))(qubits[0], qubits[1]))
        ], [{'alpha': 0.123, 'beta': 0.456, 'gamma': 0.789}]
```

jaeyoo (changes requested):
LGTM with minor stuffs.

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
jaeyoo: OR if we don't have any plan to use parameter-shift for adjoint gradient, please ignore this :)

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
```diff
@@ -0,0 +1,403 @@
...
+    output_tensor.setZero();
+
+    // Cross reference with standard google cloud compute instances
+    // Memory ~= 2 * num_threads * (2 * 64 * 2 ** num_qubits in circuits)
+    // e2s2 = 2 CPU, 8GB -> Can safely do 25 since Memory = 4GB
+    // e2s4 = 4 CPU, 16GB -> Can safely do 25 since Memory = 8GB
+    // ...
+    // This method creates 3 big state vectors per thread so reducing size
+    // here slightly.
+    if (max_num_qubits >= 25 || programs.size() == 1) {
```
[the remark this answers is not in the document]
MichaelBroughton: This is intended yes. We want each CPU thread to have < 1GB of memory for the small cases. WIth expectation calculation we only use 2 stave vectors, but here we use 3. So I cut down the number from 26 to 25.

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
```diff
@@ -0,0 +1,403 @@
...
+  void Compute(tensorflow::OpKernelContext* context) override {
+    // TODO (mbbrough): add more dimension checks for other inputs here.
+    const int num_inputs = context->num_inputs();
+    OP_REQUIRES(context, num_inputs == 5,
+                tensorflow::errors::InvalidArgument(absl::StrCat(
+                    "Expected 5 inputs, got ", num_inputs, " inputs.")));
+
+    // Create the output Tensor.
+    const int output_dim_batch_size = context->input(0).dim_size(0);
+    const int output_dim_op_size = context->input(2).dim_size(1);
```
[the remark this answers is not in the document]
MichaelBroughton: Oops, updated it.

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
```diff
@@ -0,0 +1,403 @@
...
+          if (j == partial_fused_circuits[i].size() - 1) {
+            break;
+          }
+          qsim::ApplyGate(
+              sim, qsim_circuits[i].gates[gradient_gates[i][j].index], sv);
+        }
+
+        // sv now contains psi
+        // scratch contains (sum_i paulis_sums[i] * downstream_grads[i])|psi>
+        // scratch2 now contains psi as well.
```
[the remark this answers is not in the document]
MichaelBroughton: Ahhh yes this is much more clear , changed.

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
MichaelBroughton: I don't think we can get rid of the `CopyState`. The code inside this for loop runs over the number of gradient_gate objects we have for a single gate. For a gate like FSIM which has 2 parameters, we must apply the gradient gate for the first parameter, compute the inner product, undo, apply the gradient gate for the second parameter, undo and then we can continue. This `for (int k = 0; k < gradient_gates[i][j - 1].grad_gates.size(); k++)` loop might run twice, and when it does that is when we need the copy.... _I think_ . Does this make sense ?

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
MichaelBroughton: This could definitely help others reading over it. Added.

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
MichaelBroughton: Done.

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
MichaelBroughton: Changed.

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
MichaelBroughton: See discussion above :)

tensorflow_quantum/core/ops/tfq_adj_grad_op_test.py
MichaelBroughton: Ahhh yes, will add. Thank you.

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
MichaelBroughton: Added.

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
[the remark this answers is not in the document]
MichaelBroughton: We are actually doing: `sum_j pauli_sums[i][j] * downstream_grads[i][j] |psi>`

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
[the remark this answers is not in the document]
jaeyoo: I see thanks :D

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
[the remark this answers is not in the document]
jaeyoo: Good point. thanks.

tensorflow_quantum/core/ops/tfq_adj_grad_op.cc
jaeyoo: Yes it does make sense, but I am just talking about the loop variance... Sorry, could you please help me to understand the code more?
Here is the original code with my comments on the loop variance.
```cc
        for (int j = partial_fused_circuits[i].size() - 1; j >= 0; j--) {
          for (int k = partial_fused_circuits[i][j].size() - 1; k >= 0; k--) { /* initialize sv, scratch */ }
          if (j == 0) break;

          ApplyGateDagger(
              sim, qsim_circuits[i].gates[gradient_gates[i][j - 1].index], sv); 
          /* Let `sv_before` == `sv`, `scratch_before` == `scratch` before this for-loop with k */
          for (int k = 0; k < gradient_gates[i][j - 1].grad_gates.size(); k++) {
            ss.CopyState(sv, scratch2); // sv == scratch2 == `sv_before` always due to loop invariance.
            qsim::ApplyGate(sim, gradient_gates[i][j - 1].grad_gates[k], sv); // this sv is ahead of `sv_before` with gradient_gates[i][j].grad_gates[k]. let `sv_after`
            // don't need not-found check since this is done upstream already.
            const auto it = maps[i].find(gradient_gates[i][j - 1].params[k]);
            const int loc = it->second.first;
            (*output_tensor)(i, loc) += ss.RealInnerProduct(sv, scratch) +
                                        ss.RealInnerProduct(scratch, sv); // scratch == scratch_before always. sv == sv_after
            ss.CopyState(scratch2, sv); // this undo sv's latest ApplyGate() above inside this for-loop. now sv == sv_before.
          }
          // Loop invariance : for-loop always finishes at every step with
          // sv == `sv_before`
          // scratch == `scratch_before`
          // scratch2 == `sv_before`
          ApplyGateDagger(
              sim, qsim_circuits[i].gates[gradient_gates[i][j - 1].index],
              scratch);
        }
      }
```

And here is my suggestion. To me, regardless of what gate is applied and how many of it is applied, for-loop invariance confirms that the values of `sv` and `scratch` are preserved (they are both `sv_before` and `scratch_before`), and the internal inner product values are well guaranteed with the same arguments (scratch & sv_after I mentioned). I think it also does the same undoing effect on the source `sv` for the FSIM gate with 2 parameters, because there is no change at the loop variables... 
```cc
        for (int j = partial_fused_circuits[i].size() - 1; j >= 0; j--) {
          for (int k = partial_fused_circuits[i][j].size() - 1; k >= 0; k--) { /* initialize sv, scratch */ }
          if (j == 0) break;

          ApplyGateDagger(
              sim, qsim_circuits[i].gates[gradient_gates[i][j - 1].index], sv); 
          /* say `sv_before` == `sv` `scratch_before` == `scratch` before this for-loop */
          for (int k = 0; k < gradient_gates[i][j - 1].grad_gates.size(); k++) {
            ss.CopyState(sv, scratch2); // sv == scratch2 == `sv_before` always due to loop invariance.
            qsim::ApplyGate(sim, gradient_gates[i][j - 1].grad_gates[k], scratch2); // this scratch2 is ahead of `sv_before` with gradient_gates[i][j].grad_gates[k]. let `sv_after`
            // don't need not-found check since this is done upstream already.
            const auto it = maps[i].find(gradient_gates[i][j - 1].params[k]); // regardless of what sv, scratch, scratch2 are.
            const int loc = it->second.first;
            (*output_tensor)(i, loc) += ss.RealInnerProduct(scratch2, scratch) +
                                        ss.RealInnerProduct(scratch, scratch2); // scratch == scratch_before always. scratch2 == sv_after
             // sv == sv_before without additional ss.CopyState.
          }
          // Loop invariance : for-loop always finishes at every step with
          // sv == `sv_before`
          // scratch == `scratch_before`
          // scratch2 == `sv_after` at the latest gradient_gate, but not used anywhere because it is overwritten by the `sv_before` of the next step due to `ss.CopyState(sv, scratch2)`
          ApplyGateDagger(
              sim, qsim_circuits[i].gates[gradient_gates[i][j - 1].index],
              scratch);
        }
      }
```

jaeyoo: Thanks for discussion online!

jaeyoo (approved):
LGTM!

PR #365, opened 2020-08-26
Adjoint 5/n: Added tfq.differentiators.Adjoint.

### LINKED ISSUE #256: Adjoint Differentiation

MichaelBroughton:
Adds the adjoint differentiator to `tfq.differentiators` module. Fixes #256 .

As a side note: I also added in some missing differentiator dependencies to our wheel BUILD file. It wasn't affecting wheel builds since bazel seemed to pick the missing deps up automatically, but this does give a little more peace of mind.


jaeyoo (approved):
LGTM

PR #370, opened 2020-09-01
Added locking ability to unitary op.

### LINKED ISSUE #364: Consolidate _GLOBAL_OP_LOCK
Right now `_GLOBAL_OP_LOCK` lives in `circuit_execution_ops.py`, but it does limit certain other ops from gaining access to it in appropriate way (like the unitary op for example). We should probably move this up into `quantum_context.py`

MichaelBroughton:
Fixes #364

### DIFF
--- a/tensorflow_quantum/core/ops/tfq_unitary_op.py
+++ b/tensorflow_quantum/core/ops/tfq_unitary_op.py
@@ -47,8 +47,23 @@ def get_unitary_op(
             chip.
 
     Returns:
-        An op that calculates the unitary matrix for the given programs symbol
-        names and symbol values input tensors.
+        A `callable` with the following signature:
+        ```op(programs, symbol_names, symbol_values)```
+        programs: `tf.Tensor` of strings with shape [batch_size] containing
+            the string representations of the circuits to be executed.
+        symbol_names: `tf.Tensor` of strings with shape [n_params], which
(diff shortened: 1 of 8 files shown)


tensorflow_quantum/core/ops/tfq_unitary_op.py
```diff
@@ -16,42 +16,51 @@
...
+        quantum_concurrent: Optional Python `bool`. True indicates that the
+            returned op should not block graph level parallelism on itself when
+            executing. False indicates that graph level parallelism on itself
+            should be blocked. Defaults to value specified in
+            `tfq.get_quantum_concurrent_op_mode` which defaults to True
+            (no blocking). This flag is only needed for advanced users when
+            using TFQ for very large simulations, or when running on a real
+            chip.
 
     Returns:
```
zaqqwerty, line 49: I think it would be good to expand the docstring to where it is like that for the [other ops](https://github.com/tensorflow/quantum/blob/67faf64d9e984acb29330ccd6bebe0acb463ae47/tensorflow_quantum/core/ops/circuit_execution_ops.py#L93), I find those explicit arg and shape descriptions helpful.

MichaelBroughton: Oops, good call. Added.

PR #374, opened 2020-09-03
Speed up adjoint.

MichaelBroughton:
Modified Adjoint method to do Full fusing in the forward pass instead of the same (potentially slower) partial fusing that we need to do in the backward pass.

Also changed some `>=` to `>` that were accidentally triggering re-allocations when they weren't needed.


zaqqwerty (approved):
LGTM

PR #375, opened 2020-09-07
Moved unitary onto new qsim.

MichaelBroughton:
This took a lot longer than expected.

Moved `tfq.calculate_unitary` off of old qsim. I opted to not provide a `ComputeSmall` and `ComputeLarge` option in this case, since memory requirements (most times) would put it in the `ComputeLarge` category.

This will get point number 7 on #263 .


zaqqwerty (approved):
LGTM

PR #376, opened 2020-09-08
Expanded differentiator tests.

MichaelBroughton:
Currently we don't have a great set of tests, making sure that sampled differentiators are working correctly. Added tests here for `sampled_op` behavior.

I think we should try and get this in before #345 so that we can feel a little more confident about those bigger changes coming in.

### DIFF
--- a/tensorflow_quantum/python/differentiators/gradient_test.py
+++ b/tensorflow_quantum/python/differentiators/gradient_test.py
@@ -27,7 +27,7 @@ from tensorflow_quantum.python.differentiators import linear_combination
 from tensorflow_quantum.python.differentiators import parameter_shift
 from tensorflow_quantum.core.ops import circuit_execution_ops, batch_util
 
-DIFFS = [
+ANALYTIC_DIFFS = [
     linear_combination.ForwardDifference(grid_spacing=0.0001),
     linear_combination.ForwardDifference(error_order=2, grid_spacing=0.0001),
     linear_combination.CentralDifference(grid_spacing=0.0001),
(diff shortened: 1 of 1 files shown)


tensorflow_quantum/python/differentiators/gradient_test.py
```diff
@@ -220,5 +236,97 @@ def test_analytic_value_with_simple_circuit(self, differentiator, op):
...
+        # Calculate tfq gradient.
+        symbol_values_tensor = tf.convert_to_tensor(symbol_values_array)
+        with tf.GradientTape() as g:
+            g.watch(symbol_values_tensor)
+            expectations = op(circuit, tf.convert_to_tensor(['alpha']),
+                              symbol_values_tensor, psums,
+                              tf.convert_to_tensor([[num_samples]]))
+        grads = g.gradient(expectations, symbol_values_tensor)
+        ground_truth_grads = np.array([[-1.1839752]])
+        self.assertAllClose(ground_truth_grads, grads, rtol=0.2, atol=0.2)
```
zaqqwerty, line 270: These tolerances seem a bit too large to me, could the `rtol` be removed and the `atol` be tightened up to something like `0.02`?

tensorflow_quantum/python/differentiators/gradient_test.py
```diff
@@ -27,21 +27,37 @@
...
 ]
 
-OPS = [
+SAMPLED_DIFFS = [
+    linear_combination.ForwardDifference(grid_spacing=0.1),
+    linear_combination.CentralDifference(grid_spacing=0.1),
+    parameter_shift.ParameterShift(),
+]
+
+SAMPLED_DIFFS_TOLS = [0.5, 0.5, 0.2]
```
zaqqwerty, line 44: Here also the tolerances seem a bit loose, could they be made tighter to like `1/10` as large?

tensorflow_quantum/python/differentiators/gradient_test.py
```diff
@@ -220,5 +236,97 @@ def test_analytic_value_with_simple_circuit(self, differentiator, op):
...
+                tf.convert_to_tensor([[num_samples] * n_ops] * n_programs))
+        tfq_grads = g.gradient(expectations, symbol_values_tensor)
+
+        # calculate gradients in cirq using a very simple forward differencing
+        # scheme
+        cirq_grads = _cirq_simple_finite_difference(circuit_batch,
+                                                    resolver_batch,
+                                                    symbol_names, psums)
+
+        self.assertAllClose(cirq_grads, tfq_grads, rtol=tol, atol=tol)
```
zaqqwerty, line 328: I think `rtol` should be removed, for random circuits gradients could get near zero and then `rtol` might be flagging things unnecessarily

zaqqwerty (changes requested):
Looks good, just wonder about tightening up the tolerances

MichaelBroughton:
Believe it or not, given the number of samples we are taking , this is about as tight as I can make things. We could up the number of samples, but that would blow up the test time pretty substantially. I'm worried if I make things any tighter the test would get flaky.

zaqqwerty (approved):
Makes sense, LGTM

PR #378, opened 2020-09-09
Final removal of dead qsim code.

### LINKED ISSUE #263: [Tracking] qsim migration

MichaelBroughton:
It was a long road, but we are finally done with #263 . This marks the final removal of all of the dead qsim code. With this change we will no longer have our dead code lying around and will receive performance updates as they happen over at: https://github.com/quantumlib/qsim . Maybe even CUDA one day soon ;)

Fixes #263 .


zaqqwerty (approved):
Searched through the code in this branch and I agree dead code is gone now, LGTM

PR #379, opened 2020-09-11
Consolidate differentiators

zaqqwerty:
This PR builds on #345.  That PR introduced a new interface for defining differentiators.  This PR updates the base class `Differentiator` to depend on `get_intermediate_logic`, thus removing the need for inheritors to implement `differentiate_analytic` or `differentiate_sampled`.  Thus this PR also removes those functions from `LinearCombination` and `ParameterShift`.


peezoslug (approved):

peezoslug (approved):

zaqqwerty:
Will be done with the new interface instead.

Issue #380, opened 2020-09-11
[Discussion] Adding quantum math features


MichaelBroughton:
Sometimes, it is useful to be able to analyze circuits in ways that aren't valid on a quantum computer, but might still be interesting for theory reasons. I came across this issue a little while ago when wanting to compute the overlap `<psi_1 | psi_2>` of a state and had a hard time keeping memory limits under control with larger batch and state sizes using `tfq.layers.State`. I ultimately ended up simulating the circuits one by one, and doing the inner products in tf manually. It worked, but it wasn't pretty.

Would people be interested in implementing things like:

**Examples that can be done right now, but are tricky to get right without blowing up memory:**
```python3
tfq.math.overlap # => analytically compute state overlap between the two circuits
tfq.math.amplitudes # => analytically compute requested state amplitudes (could use qsimh ?).
tfq.math.fideltiy # => analytically compute fidelity between two circuits.
```

**Other interesting math functions might be worth having**
```python3
tfq.math.operator_matrix # => ingest a cirq.PauliSum and compute the matrix
tfq.math.operator_commutators # => compute commutators on operator matrices
```

Does anyone else have any suggestions for useful math function ? @dabacon @zaqqwerty @jaeyoo @we-taper 

Final thought: If we do end up making this `.math` module , maybe we should move `calculate_unitary` and `calculate_state` inside of it.

jaeyoo:
```
tfq.math.fubini_study_metric # -> used similar to fidelity
tfq.math.fisher_information_matrix # -> used in calculating Hessian
tfq.math.relative_entropy # -> used in measure between two mixed states.
```

therooler:
To add to my comments in the TFQ sync meeting about how useful a `tfq.math.fubini_study_metric` would be:

This is what I had to do in TensorFlow 1.x to calculate the Quantum Natural Gradient:
```python
def QuantumNaturalGradient(state, loss, vrs, optimizer=None, stability_shift=None, learning_rate: float = 0.01):
    """
    Implementation of the quantum natural gradient gradient method

    Args:
        *state (Tensor)*:
            Output state of the circuit.

        *loss (Tensor)*:
            Loss function that uses the provided state.

        *vrs (Variables)*:
            Variables to be optimized.

        *optimizer (Tensorflow Optimizer)*:
            Tensorflow optimizer from the tf.train API

    Returns (Operation):
        Train step operation.

    """
    if optimizer == None:
[...]

PR #387, opened 2020-09-17
Inner product

MichaelBroughton:
Adds the `tfq.math` module. First op in is `inner_product`. For reference #380 .

Still TODO: Write docstring code snippet.


jaeyoo (approved):
LGTM

PR #393, opened 2020-09-19
TF Quantum Starter Example

Rishit-dagli:
Add a TF Quantum example for beginners to get started with TF Quantum. The notebook demonstrates the following pretty simple operations-

- Placing a Qubit
- Working with simple quantum data
- Building a quantum circuit
- Creating and training a hybrid model 
- Inferencing the hybrid model

I am aware that the tutorial [hello_many_worlds](https://github.com/tensorflow/quantum/blob/master/docs/tutorials/hello_many_worlds.ipynb) exists which is a beginner tutorial however, this tutorial has different aims.

This PR was originally created on tensorflow/examples#244 ported to this repo as suggested by @lamberta .


zaqqwerty:
So most of this code looks to be copied from our already existing notebook at https://github.com/tensorflow/quantum/blob/research/binary_classifier/binary_classifier.ipynb
I see that, relative to the existing notebook, you simplified the quantum data input to have just two entries.  Perhaps if there is interest in the existing notebook becoming a tutorial we could consider such a simplification. @MichaelBroughton thoughts on this?

Rishit-dagli:
> So most of this code looks to be copied from our already existing notebook at https://github.com/tensorflow/quantum/blob/research/binary_classifier/binary_classifier.ipynb
> I see that, relative to the existing notebook, you simplified the quantum data input to have just two entries. Perhaps if there is interest in the existing notebook becoming a tutorial we could consider such a simplification. @MichaelBroughton thoughts on this? (see also page 9 of the [TFQ paper](https://arxiv.org/pdf/2003.02989.pdf))

That's right I built this as a simplification to the existing one, to make it a lot easier for beginners to understand.

MichaelBroughton:
I think we might be able to fit this in as a tutorial, right after the hello many worlds and before the MNIST classification tutorial. The goal of our tutorials is to showcase the functionality and features of the library so that experts have an easy time making the mental connections between how TFQ functions and QC theory (which we generally expect people to have a strong grasp of). 

Even though this is a more "beginner" example and doesn't touch any new library code that hello many worlds doesn't already touch, I think showcasing good visualizations of the single qubit rotation problem might give experts a little more opportunity to connect with and understand how TFQ works.

@lamberta Are you alright if I give this a review and then hand off to you for final say ?

MichaelBroughton (changes requested):
A couple of thematic and large changes I think we need here, before we go into a more serious review would be:

1. If this is going to be a more beginner oriented tutorial we should be sure to include the appropriate descriptions of all the processes in the notebook and do our best to explain every step of what is going on. Throughout the notebook things are a little bit unclear. Ideally after the user is done reading the tutorial, they shouldn't have many questions like: Why are we placing a qubit and rotating it in this way ? Why is the shape of the labels array [2,2] ? what does a label of 1 vs 0 mean ? An effort should be made to leave the user with as few of these kinds of questions as possible (since this is a beginner tutorial this is going to be especially tricky and important).

2. The length of this tutorial is very short. We create the circuit, generate the data, train the model show a graph and then we are done. To help beginners we should definitely showcase some more math and make sure we firm up understanding for them at every step. For example, a good spot in the tutorial I think would benefit from having more detail is the data preparation section. Instead of making 2 datapoints, make 2,000 and then visualize them like we did in the research tutorial this is based on. You could also show some math like "here are the polar coordinates for this particular datapoint. You can see that because the qubit is rotated in this way it's label is this...." etc.

Overall, this is a good first step, but we will need to build things up a fair bit more if we are going to get it ready to showcase as a tutorial.

lamberta:
Thanks, @MichaelBroughton 
I agree. A beginner's tutorial is good, but need to be mindful of the TFQ audience: beginning researchers (and not beginning practitioners). Seems better to port https://github.com/tensorflow/quantum/blob/research/binary_classifier/binary_classifier.ipynb with the sections added here, but with added explanation.

Rishit-dagli:
Thanks, @MichaelBroughton These points make sense and I will start working on them.

1. I am just a bit unclear about adding descriptions, I can most certainly have a description to address the questions you mentioned in the comment but would anything more be required to make this more clear?
2. This makes sense, I will definitely showcase some more math and make sure I firm up understanding for them.

Thanks, @lamberta 
> A beginner's tutorial is good, but need to be mindful of the TFQ audience: beginning researchers (and not beginning practitioners)
This is something I need to be mindful of while making changes and maybe could change the tone of the tutorial since this tutorial was originally made for a talk I gave at a TensorFlow User Group where some folks were practitioners and was made to help them out too.

mhucka:
@Rishit-dagli Thank you for your work on this. I realize this PR is quite old now; nevertheless, I wanted to check if you are still interested in finish this. It sounds like it would be a useful addition.

PR #399, opened 2020-09-30
Removed unused release files.

MichaelBroughton:
These files have all moved inside of google and changed a great deal in order to do our wheel builds. We won't continue to maintain them on the outside so let's just remove them.


zaqqwerty (approved):
LGTM

PR #404, opened 2020-10-06
Master

peezoslug:


mihaimaruseac:
Hi @peezoslug,

It seems you keep creating PRs with no explanation for what they are doing, both in TF and in TF Quantum (at least). Please note that spam PRs are considered bad both by the projects' code of conduct / contribution guidelines as well as by Hacktoberfest organizers (in case these PRs are for Hacktoberfest). Continuous spam can lead to you being banned.

PR #406, opened 2020-10-06
Added rectangular TFIM data and tests

therooler:
In this next installment of the the quantum data set endeavor, the Transverse Field Ising model on a 3x3, 3x4 and 4x4 lattice mapped to a torus! How exciting.

Most stuff here is the same as before. One thing we need to decide is how to define the (sub)graphs we need for the circuit definition. Right now I just hardcoded the (sub)graph of the circuit as 
```python
edges = {
                'g1': [[0, 3], [1, 2], [4, 7], [5, 8], [6, 9], [10, 11]],
                'g2': [[0, 1], [2, 5], [3, 4], [6, 7], [8, 11], [9, 10]],
                'g3': [[0, 9], [1, 10], [2, 11], [3, 6], [4, 5], [7, 8]],
                'g4': [[0, 2], [1, 4], [3, 5], [6, 8], [7, 10], [9, 11]]
            }
```
which is necessary because we parameterize these 4 subgraphs individually. I stored the data again at https://github.com/therooler/tfq_data, so if you add this to the public google storage, the test should pass. (I'll update the docstrings tomorrow). 

Once we converge on a way to handle these 2D models in a nice way, I can also code up the TFIM for the 12 site Kagome lattice. The data for that is already in my repo, and I'm planning to add the 18 site data as well. Finally, I am in the process of gathering and checking the data for the 'closed' boundary conditions for the 1D chains we have now, so that should be uploaded to the repo soon as well.


MichaelBroughton:
Awesome! Try linking up here: https://storage.googleapis.com/download.tensorflow.org/data/quantum/spin_systems/TFI_rect.zip

Then we can get started on the review :)

therooler:
Tests passed!

tensorflow_quantum/datasets/spin_system.py
```diff
@@ -557,3 +557,264 @@ def xxz_chain(qubits, boundary_condition="closed", data_dir=None):
             additional_info)))
 
     return resolved_circuits, labels, hamiltonians, additional_info
+
+
+def tfi_rectangular(qubits, boundary_condition="torus", data_dir=None):
+    """Transverse field Ising-model quantum data set.
```
MichaelBroughton, line 563: Might want to specify it's 2D to seperate things from the `tfi_chain`

MichaelBroughton (changes requested):
Looking really good! Would you be able to revert the changes made in the beginning of `spin_system_test.py`, it looks like we've accidentally reverted from things like `final_state_vector` back to the deprecated `final_state` etc.

tensorflow_quantum/datasets/spin_system_test.py
```diff
@@ -83,7 +83,7 @@ def test_paulisum(self):
             }
             for n in self.random_subset_tfi_chain:
                 phi = cirq.Simulator().simulate(circuits[n]).final_state_vector
-                e = pauli_sums[n].expectation_from_state_vector(phi, qubit_map)
```
MichaelBroughton, line 86: Still missing this deprecation :)

tensorflow_quantum/datasets/spin_system_test.py
```diff
@@ -182,7 +182,7 @@ def test_paulisum(self):
             }
             for n in self.random_subset_xxz_chain:
                 phi = cirq.Simulator().simulate(circuits[n]).final_state_vector
-                e = pauli_sums[n].expectation_from_state_vector(phi, qubit_map)
```
MichaelBroughton, line 185: and here.

tensorflow_quantum/datasets/spin_system_test.py
```diff
@@ -217,6 +217,108 @@ def test_param_resolver(self):
                                     rtol=1e-3)
 
 
+class TFIRectangularTest(tf.test.TestCase):
+    """Testing tfi_rectangular."""
+    # pylint: disable=C0103
```
MichaelBroughton, line 222: What exactly is this disabling ?

MichaelBroughton (commented):
Looking good, let's get rid of the last of those deprecation warnings, then we are good to merge.

tensorflow_quantum/datasets/spin_system_test.py
therooler: Overloading the `setUpClass` method is making pylint sad because the method name is not lowercase. C0103 allows me to pass the pylint checks.

MichaelBroughton (approved):

PR #409, opened 2020-10-08
Gradient circuits 1/n: interface

zaqqwerty:
Having finished discussions in the design doc, I am moving to implement the new gradient circuits feature.  To ease review I am splitting the implementation into a few pieces.  Here, I add the interface for `get_gradient_circuits` to the base `Differentiator` class.


docs/tutorials/gradients.ipynb
```diff
@@ -626,6 +626,12 @@
         "        pass\n",
         "\n",
         "    @tf.function\n",
+        "    def get_gradient_circuits(self, programs, symbol_names, symbol_values):\n",
+        "        \"\"\"See base class description.\"\"\"\n",
+        "        raise NotImplementedError(\n",
+        "            \"Gradient circuits are not implemented in this tutorial.\")\n",
```
MichaelBroughton, line 632: Should probably provide a meaninful description here other than "See base class".

tensorflow_quantum/python/differentiators/differentiator.py
```diff
@@ -179,6 +179,86 @@ def refresh(self):
...
+        computer, where the objective function resides, can only access
+        bitstrings sampled from states produced on the QPU; derivatives must be
+        computed from these samples instead. This function exposes the circuits,
+        symbols, and values required for computing such gradients on real QPUs.
+
+        NOTE: most users should simply hand a differentiator to the circuit
+        executor of their choice (for example, see the docs for
+        `tfq.layers.Expectation`, which accepts a `differentiator` initializer).
+        This function is intended for advanced users who need more flexibility
+        than the standard workflow allows.
```
MichaelBroughton, line 209: This is kind of redundant. Having seen our taking gradient tutorial or being an expert on QC most people should already know this. What do we think of this description:

```python3
@abc.abstractmethod
def get_gradient_circuits(self, programs, symbol_names, symbol_values):
"""Return the circuits used to compute gradients for a given set of forward pass circuits.

Prepares (but does not execute) all intermediate circuits needed to calculate the gradients for the
given forward pass circuits specified by `programs`, `symbol_names` and `symbol_values`. The returned
`tf.Tensor` objects give all necessary information to recreate the internal logic of the differentiator with
finer grained control over the calculation process. <insert huge formula here>. This can be useful for advanced users when executing on QPU devices or defining custom differentiation methods.

Note: This feature is intended for advanced users who want more fine control during graph construction and
execution. 

"""
```

MichaelBroughton (changes requested):
Looking good. The description of the function reads a bit like an academic introduction to the math. Might be good if we just moved to more of a "It does what it says on the box" kind of approach and trim things down a bit. What do you think ?

tensorflow_quantum/python/differentiators/differentiator.py
zaqqwerty: I like this, describes what the function more efficiently.

docs/tutorials/gradients.ipynb
zaqqwerty: Expanded the docstring

zaqqwerty:
Pared down the docstring for `get_gradient_circuits` in the base class and added some description to the gradients tutorial docstring, what do you think?

tensorflow_quantum/python/differentiators/differentiator.py
```diff
@@ -179,6 +179,71 @@ def refresh(self):
...
+        batch) with [n_vals] output values per function, then the derivative of
+        the computation will have shape [batch_size, n_params, n_vals]. In terms
+        of the Args and Returns of `get_gradient_circuits`, each entry of this
+        derivative can then be computed as
+
+        d(<programs[i](symbol_values[i])|
+            f[i]|programs[i](symbol_values[i])>)
+        / d(symbol_names[k]) = sum_m (batch_mapper[i, k, m] *
+            (<batch_programs[i, m](batch_symbol_values[i, m])|
+                f[i]|batch_programs[i, m](batch_symbol_values[i, m])>)
```
MichaelBroughton, line 202: This notation is kind of confusing. What exactly do you mean when you write `f[i] | batch_programs[i, m]` ?

Would it be easier to just explain: "This is how the returned values can be used to reconstruct the original gradients" ? I feel like with that explanation a user can just pick apart whichever component they want and develop a better understanding of what is happening. 

Thoughts ?

zaqqwerty: `|batch_programs[i, m](batch_symbol_values[i, m])>` is the ket representing the state generated by program `batch_programs[i, m]` using the parameter values `batch_symbol_values[i, m]`.  Thus `<batch_programs[i, m](batch_symbol_values[i, m])|f[i]|batch_programs[i, m](batch_symbol_values[i, m])>` is the expectation value of function `f[i]` against that state.  I suppose if the notation is not clear enough here it won't be for other people either though.

Would the docstring be explicit enough for a user to combine the outputs appropriately if there were not an equation?  I'm deep into this enough that I can't quite tell, if it is I am fine to leave it out (this equation was meant to be the same as the one you commented on in the design doc).

An example of the list of vector-valued functions `f` is the PauliSums tensor used by the other ops.

zaqqwerty: Oh also I realized that `f[i] | batch_programs[i, m]` starting a line is just due to line breaking for the 80 char restriction, if that is what makes it confusing I could stop the break and add a pylint ignore

MichaelBroughton: What about just writing a minimal running snippet that combines the outputs to compute a gradient ? Would that work ?

zaqqwerty: Good point, that sounds better, I'll add a snippet that actually does it instead of writing up the math

zaqqwerty:
Added some code examples, is it what you had in mind?

MichaelBroughton:
This snippet is good and might be something worth adding to a tutorial, but as it stands it's pretty complicated and touches a lot of areas of the library. I was thinking a simpler snippet with something like:

```python3
<Create a differentatioator> a
... = a.get_gradient_circuits(...)
manual_results = <math to get same result as differentiate_analytic>
automatic_results = a.differentiate_analytic(...)
manual_results == automatic_results
True

```

zaqqwerty:
Like this?
```python

        Suppose we have some inputs `programs`, `symbol_names`, and                                                                                                                                                                                                                
        `symbol_values`.  To get the derivative of the expectation values of a                                                                                                                                                                                                     
        tensor of PauliSums `pauli_sums` with respect to these inputs, for some                                                                                                                                                                                                    
        program index `i` and symbol index `k`, do:                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                   
        >>> diff = <some differentiator>()                                                                                                                                                                                                                                         
        >>> (                                                                                                                                                                                                                                                                      
        ...     batch_programs, new_symbol_names, batch_symbol_values,                                                                                                                                                                                                             
        ...     batch_mapper                                                                                                                                                                                                                                                       
        ... ) = diff.get_gradient_circuits(                                                                                                                                                                                                                                        
        ...     programs, symbol_names, symbol_values)                                                                                                                                                                                                                             
        >>> d_programs_i_d_symbol_k_manual = tf.reduce_sum(                                                                                                                                                                                                                        
        ...     batch_mapper[i, k] * tfq.layers.Expectation()(                                                                                                                                                                                                                     
        ...         batch_programs[i], symbol_names=new_symbol_names,                                                                                                                                                                                                              
        ...         symbol_values=batch_symbol_values[i],                                                                                                                                                                                                                          
        ...         operators=tf.tile(tf.expand_dims(pauli_sums[i], 0),                                                                                                                                                                                                            
        ...             [tf.shape(batch_programs), 1])), 1)                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                   
        This has the same effect as taking the gradient of Expectation layer:                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                   
        >>> with tf.GradientTape() as g:                                                                                                                                                                                                                                           
        >>>     symbol_values =                                                                                                                                                                                                                                                    
        >>>     g.watch(symbol_values)                                                                                                                                                                                                                                             
        >>>     exact_outputs = tfq.layers.Expectation()(                                                                                                                                                                                                                          
        ...         programs[i], symbol_names=symbol_names,                                                                                                                                                                                                                        
        ...         symbol_values=symbol_values, operators=pauli_sums[i])                                                                                                                                                                                                          
        >>> d_programs_i_d_symbol_k_auto = g.gradient(                                                                                                                                                                                                                             
        ...     exact_outputs, symbol_values)                                                                                                                                                                                                                                      
        >>> all(d_programs_i_d_symbol_k_auto == d_programs_i_d_symbol_k_manual)
        True
```

MichaelBroughton:
Yeah I think this looks great and would let anyone interested in finer grained control understand what is going on.

This snippet is indexing right into the tensors which limits this snippet to only working in eager mode. Would it be possible to have just one call to `tfq.layers.Expectation` and not index into any of the tensors and get the gradient for the full batch instead of just the one circuit ?

zaqqwerty:
Added the following docstring:
```python
        Suppose we have some inputs `programs`, `symbol_names`, and
        `symbol_values`.  To get the derivative of the expectation values of a
        tensor of PauliSums `pauli_sums` with respect to these inputs, do:
        >>> diff = <some differentiator>()
        >>> (
        ...     batch_programs, new_symbol_names, batch_symbol_values,
        ...     batch_mapper
        ... ) = diff.get_gradient_circuits(
        ...     programs, symbol_names, symbol_values)
        >>> exp_layer = tfq.layers.Expectation()
        >>> batch_pauli_sums = tf.tile(
        ...     tf.expand_dims(pauli_sums, 0),
        ...     [1, tf.shape(batch_mapper)[2], 1])
        >>> batch_expectations = tf.map_fn(
        ...     lambda x: exp_layer(x[0], new_symbol_names, x[2], x[3]),
        ...     (batch_programs, batch_symbol_values, batch_pauli_sums),
        ...     fn_output_signature=tf.float32)
        >>> grad_manual = tf.reduce_sum(tf.map_fn(
        ...     lambda x: tf.reduce_sum(x[0] * x[1], -1),
        ...     (batch_mapper, batch_expectations),
        ...     fn_output_signature=tf.float32), -1)
        For best performance the user should flatten the inputs and feed them
        to the Expectation layer all at once, but the above better illustrates
        the structure of the computation. To perform the same gradient
        calculation automatically:
        >>> with tf.GradientTape() as g:
        >>>     g.watch(symbol_values)
        >>>     exact_outputs = tfq.layers.Expectation()(
        ...         programs, symbol_names=symbol_names,
        ...         symbol_values=symbol_values, operators=pauli_sums)
        >>> grad_auto = g.gradient(exact_outputs, symbol_values)
        >>> all(derivative_manual == derivative_auto)
        True
"""

**Comment by MichaelBroughton** [comment]
This snippet still has multiple calls to expectation. From the `map_fn` snippet:

> "map_fn unstacks elems on axis 0 to obtain a sequence of elements; calls fn to transform each element; and then stacks the transformed values back together."

This means we will still wind up with many calls to expectation which is fine, but not as fast as it could be. What about doing something like this instead:

```python3
batch_expectation = tfq.layers.Expectation()(
    tf.reshape(batch_programs, [-1, <rest of shape excluding first dim>]),
    new_symbol_names,
    tf.reshape(batch_symbol_values, [-1, <rest of shape excluding first dim>]),
    tf.reshape(batch_pauli_sums, [-1, <rest of shape excluding first_dim>] ))

batch_expectations = tf.reshape(batch_expectations, ...)
grad_manual = tf.einsum('...', batch_mapper, batch_expectations)
```

This way we only get one call and we can get rid of uneeded calls to `map_fn` for `batch_expectation` and for `grad_manual`. Would that work ?

zaqqwerty:
@MichaelBroughton I updated the docstring code to flatten the programs and use einsum

MichaelBroughton (approved):

PR #416, opened 2020-10-22
Remove stale checker.

MichaelBroughton:
I think this stale checker wound up being more trouble that it was worth. Anyone have any strong feelings ?


zaqqwerty (approved):
LGTM.  Seems we have a low enough rate of issues and PRs that managing them by hand is easy enough.

PR #419, opened 2020-10-26
Add docs for append_circuit op

### LINKED ISSUE #417: Op append_circuit not showing up in docs
The component `append_circuit` is available in the codebase but doesn't show up in the docsite.  I think this can be resolved by making a python function around the op as was done for `resolve_parameters` [here](https://github.com/tensorflow/quantum/blob/a2c02bad57a83cf0c149d322a691b184e31c413f/tensorflow_quantum/core/ops/tfq_utility_ops.py#L63).

zaqqwerty:
Resolves #417


MichaelBroughton (approved):
Thanks for catching this! Looks good!

zaqqwerty:
Could you merge this in? (I don't have write access)

Issue #361, opened 2020-08-24
Fix compiler warnings


MichaelBroughton:
When compiling our C++ code we are starting to get a lot of (mostly) harmless warnings about comparison between unsigned and signed types etc. It would be nice if we could comb through these and make the tweaks to make them go away. This isn't an urgent issue, but it would be nice to clean up the clutter.

mhucka:
This has been pretty much completed via these PRs:
- #839 
- #838 
- #837 
- #836 
- #835 
- #834

PR #424, opened 2020-11-03
remove some error messages due to sign/unsigned comparisons

tonybruguier:
Very first try for issue #361

### DIFF
--- a/tensorflow_quantum/core/src/util_qsim.h
+++ b/tensorflow_quantum/core/src/util_qsim.h
@@ -45,7 +45,7 @@ struct QsimFor {
   template <typename Function, typename... Args>
   void Run(uint64_t size, Function&& func, Args&&... args) const {
     auto worker_f = [&func, &args...](int64_t start, int64_t end) {
-      for (uint64_t i = start; i < end; i++) {
+      for (int64_t i = start; i < end; i++) {
         // First two arguments in RUN appear to be unused.
         std::forward<Function>(func)(-10, -10, i, std::forward<Args>(args)...);
       }
@@ -162,8 +162,8 @@ tensorflow::Status ComputeExpectationQsim(const tfq::proto::PauliSum& p_sum,
     }
     // copy from src to scratch.
     ss.CopyState(state, scratch);
-    for (int j = 0; j < fused_circuit.size(); j++) {
-      qsim::ApplyFusedGate(sim, fused_circuit[j], scratch);
+    for (const qsim::GateFused<QsimGate>& fused_gate : fused_circuit) {
+      qsim::ApplyFusedGate(sim, fused_gate, scratch);
     }
 
     if (!status.ok()) {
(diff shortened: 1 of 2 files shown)


tensorflow_quantum/core/src/util_qsim.h
```diff
@@ -308,7 +308,10 @@ tensorflow::Status AccumulateOperators(
   ss.CopyState(source, scratch);
   ss.SetAllZeros(dest);
 
-  for (int i = 0; i < p_sums.size(); i++) {
+  DCHECK_EQ(p_sums.size(), op_coeffs.size());
```
MichaelBroughton, line 311: Why add the `DCHECK_EQ` ?

tensorflow_quantum/core/src/util_qsim.h
```diff
@@ -308,7 +308,10 @@ tensorflow::Status AccumulateOperators(
...
+  const int num_p_sums = p_sums.size();
+
+  for (int i = 0; i < num_p_sums; i++) {
```
MichaelBroughton, line 314: Would `size_t` for `i` work instead ? It seems like that would be the type that `p_sums.size()` returns ?

MichaelBroughton (commented):
Thanks for taking this on! Left a few questions in the code, once we get those sorted, should be good to merge.

tensorflow_quantum/core/src/util_qsim.h
tonybruguier: Yes, it would work. I was just trying to follow the Google style guide:
https://google.github.io/styleguide/cppguide.html#Integer_Types

> Because of historical accident, the C++ standard also uses unsigned integers to represent the size of containers - many members of the standards body believe this to be a mistake
> and try to avoid unsigned types (except for representing bitfields or modular arithmetic)

That being said, I am far from a C++ expert and I'm quite flexible on the whole thing.

tensorflow_quantum/core/src/util_qsim.h
tonybruguier: It's just that there was an existing comment on top of the file that said:
// Assumes p_sums.size() == op_coeffs.size()

It seems to be in line with what the code does, so I figured it would be better to enforce it rather than running the risk of stepping outside an array.

tonybruguier (commented):
I replied to your comments. These are very weak preferences, and I'm more than happy to modify the code.

tensorflow_quantum/core/src/util_qsim.h
MichaelBroughton: Fair enough. Let's keep it then.

tensorflow_quantum/core/src/util_qsim.h
MichaelBroughton: The guide also says using size_t is fine and in our case it'd let us cut down to just 

```C++
for(size_t i =0;i<p_sums.size();i++){
  ...
}
```

MichaelBroughton:
Got the replies. Let's stick with `size_t` in the for loop and leave the `DCHECK_EQ` in.  Then we should be good to merge!

tensorflow_quantum/core/src/util_qsim.h
tonybruguier: Done.

tonybruguier:
Thanks. PTAL.

MichaelBroughton (approved):

PR #446, opened 2020-12-04
Add explicit use_cxx11_abi=0 to configuration.

MichaelBroughton:
Just a small quality of life improvement to allow us to write statements like 
`bazel test ...` without having to worry about adding in the long abi flag everytime.


zaqqwerty (approved):
LGTM

Issue #422, opened 2020-10-29
Add support for multi qubit controlled gates


MichaelBroughton:
Once qsim has released a new github version that contains multi qubit controlled gate support we should look into supporting over here in TFQ. This will roughly entail:

~~1. Opening a PR to upgrade from qsim 0.3.1 -> 0.6.0~~
~~2. Adding serialization primitives to `serializer.py` for multi qubit gates and corresponding tests.~~
3. Hooking up new serialization primitives in underlying C++ and other ops.

xref for adding gates: #257

@we-taper has indicated some interest in tackling this and I'll be happy to help out when I can!

PR #447, opened 2020-12-09
Serialize multiqubit

MichaelBroughton:
Addresses step 2 in #422 .

The final step is to link these gates into the new qsim simulator. Right now most of the C++ code still ignore `control_qubits` and `control_values` proto fields.

With the help of the `DelayedAssignmentGate` this change adds the `control_qubits` and `control_values` field to every gates [args](https://github.com/quantumlib/Cirq/blob/master/cirq/google/api/v2/program.proto#L110) field. Updates all previous tests to use controlled operations as well and added a test for invalid `control_qubits`. Plz review closely, I'm always a little concerned when touching code in `serializer.py` since it's used so widely.

### DIFF
--- a/tensorflow_quantum/core/serialize/serializer.py
+++ b/tensorflow_quantum/core/serialize/serializer.py
@@ -539,7 +539,8 @@ def serialize_circuit(circuit_inp):
     # Demote cirq.controlled_operations (controlled gates) to their sub_gate
     # types with _tfq_control_qubits and _tfq_control_values fields so that
     # the gates can still get picked up by the serializer. There would be no way
-    # to discern controlledgates from one another otherwise.
+    # to discern controlledgates from one another otherwise. This
+    # "momentary demotion" occurs with the help of the DelayedAssignmentGate.
     for i, moment in enumerate(circuit):
         for op in moment:
             if isinstance(op,
--- a/tensorflow_quantum/core/serialize/serializer.py
+++ b/tensorflow_quantum/core/serialize/serializer.py
@@ -153,7 +153,6 @@ def _optional_control_promote(gate, control_qubits_message,
         r = int(r)
         c = int(c)
         cv = int(cv)
-        # gate = gate.controlled_by(cirq.GridQubit(r, c), controlled_value=[cv])
         qbs.append(cirq.GridQubit(r, c))
         vals.append([cv])
 
--- a/tensorflow_quantum/core/serialize/serializer.py
+++ b/tensorflow_quantum/core/serialize/serializer.py
@@ -140,21 +140,12 @@ class DelayedAssignmentGate(cirq.Gate):
             *self._control_qubits, control_values=self._control_values)
 
 
-def _optional_control_promote(gate, control_qubits_message,
-                              control_values_message):
+def _optional_control_promote(gate, qubits_message, values_message):
     """Optionally promote to controlled gate based on serialized control msg."""
-    if control_qubits_message == '' and control_values_message == '':
+    if qubits_message == '' and values_message == '':
         return gate
-    qbs = []
-    vals = []
-    for qb, cv in zip(control_qubits_message.split(','),
-                      control_values_message.split(',')):
-        r, c = qb.split('_')
-        r = int(r)
-        c = int(c)
-        cv = int(cv)
-        qbs.append(cirq.GridQubit(r, c))
-        vals.append([cv])
+    qbs = [v2.qubit_from_proto_id(qb) for qb in qubits_message.split(',')]
+    vals = [int(cv) for cv in values_message.split(',')]
 
     return DelayedAssignmentGate(gate, qbs, vals)
 
--- a/tensorflow_quantum/core/serialize/serializer.py
+++ b/tensorflow_quantum/core/serialize/serializer.py
@@ -135,9 +135,13 @@ class DelayedAssignmentGate(cirq.Gate):
     def _qid_shape_(self):
         raise ValueError("Called qid_shape on workaround class.")
 
+    # pylint: disable=invalid-name
     def on(self, *qubits):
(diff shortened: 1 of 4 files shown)


zaqqwerty:
I'll have some time to look at this in detail tomorrow

tensorflow_quantum/core/serialize/serializer.py
```diff
@@ -434,6 +536,20 @@ def serialize_circuit(circuit_inp):
...
 
+    # Demote cirq.controlled_operations (controlled gates) to their sub_gate
+    # types with _tfq_control_qubits and _tfq_control_values fields so that
+    # the gates can still get picked up by the serializer. There would be no way
+    # to discern controlledgates from one another otherwise.
+    for i, moment in enumerate(circuit):
+        for op in moment:
+            if isinstance(op,
+                          cirq.ops.controlled_operation.ControlledOperation):
+                tfq_compatible = op.sub_operation
```
zaqqwerty, line 547: I thought maybe recursion of controlled operations could be a problem here, but I tried the following snippet:
```
q0, q1, q2 = cirq.GridQubit.rect(1, 3)
cx = cirq.X(q0).controlled_by(q1)
ccx = cirq.ops.ControlledOperation((q2,), cx)
```
and it turns out `ccx.sub_operation` is still an X gate.  Could you add a comment here, maybe with a code snippet like this, just to clarify that it is not possible to naturally create a sub operation that is Controlled?

tensorflow_quantum/core/serialize/serializer_test.py
```diff
@@ -425,15 +473,17 @@ class SerializerTest(tf.test.TestCase, parameterized.TestCase):
...
     def test_serialize_circuit_valid(self, circ_proto_pair):
         """Test conversion of cirq Circuits to tfq_gate_set proto."""
         self.assertProtoEquals(serializer.serialize_circuit(circ_proto_pair[0]),
                                circ_proto_pair[1])
 
     @parameterized.parameters([{
         'circ_proto_pair': v
-    } for v in _get_valid_circuit_proto_pairs()])
+    } for v in _get_controlled_circuit_proto_pairs() +
+                               _get_circuit_proto_pairs()])
```
zaqqwerty, line 486: I'm having trouble understanding why this test is passing.  For example, the first entry in the list returned by `_get_circuit_proto_pairs()` is an HPow gate.  `_make_controlled_circuit` should promote this to a `cirq.ControlledOperation`, and the corresponding proto is promoted to have `control_qubits` and `control_values` fields.  But deserializing that proto should yield a `DelayedAssignmentGate`; how does it come out equal to the `cirq.ControlledOperation` in `circ_proto_pair[0]`?

zaqqwerty (changes requested):
Just a couple clarifications

tensorflow_quantum/core/serialize/serializer_test.py
MichaelBroughton: The DelayedAssignmentGate is what is getting sent into the `op_serializer.py` functions. These functions will very soon call the `.on` method which the `DelayedAssignmentGate` uses to create an instance of the apropriate gate for serialization. Without it all `cirq.ControlledOperations` look the same to the serializer and parsing them would be more of a pain.

zaqqwerty: Yeah the DelayedAssignmentGate is needed to separate out the controlling qubits.  I just mean in the test below there are parameters such that `serializer.deserialize_circuit(circ_proto_pair[1])` should be a `DelayedAssignmentGate` while `circ_proto_pair[0]` should be a `cirq.ControlledOperation`, and I was wondering how these evaluate to equal

MichaelBroughton: Here is the life of a proto as it gets deserialized:

Proto -> `op_deserializer.py` -> begins looping over operations and invokes the specific deserialization rule for the current operation -> that procedure returns a gate that has not yet had qubit's chosen via a call to the `.on` method (An instance of delayedassignmentgate) -> `op_deserializer.py` calls the `.on` method and the DelayedAssignmentGate returns a `cirq.ControlledOperation` with the `sub_gate` equal to the original gate found in the proto. 

This gets called only within `op_deserializer.py`:
```python3
def on(self, *qubits):
        """Returns gate_callable on qubits controlled by contol_qubits."""
        return self._gate_callable(*qubits).controlled_by(
            *self._control_qubits, control_values=self._control_values)
```

So the DelayedAssignmentGate only exists for a moment before returning the controlledoperation that is the true form of the gate. The `DelayedAssignmentGate` class is needed so that we can assign control qubit information outside of `op_deserializer.py` since it doesn't natively support controlled gates.



Does this help clear things up ?

zaqqwerty: Thanks for the details here, I understand what's happening now: the `DelayedAssignmentGate` only exists between lines 105 and 106 of `cirq.google.op_deserializer`, after that it is the regular `ControlledOperation`; so the new class is a workaround for the existing op `from_proto` implementation.

Could you add a short addition to the DelayedAssignmentGate to clarify things for future readers, maybe something like "DelayedAssignmentGate is only ever instantiated locally by the native Cirq deserializer (inside cirq.google.GateOpDeserializer.from_proto) and then discarded"?

MichaelBroughton: Added more clarity in the comments found in the `serialize_circuit` function.

tensorflow_quantum/core/serialize/serializer.py
MichaelBroughton: This is a feature of the Cirq library so I don't think we really need to make that clear here.

zaqqwerty (approved):
LGTM

PR #448, opened 2020-12-10
Add latest TF blog posts as landing cards on the site.

MichaelBroughton:
Add in two cards to the latest blog posts. Should I put these at the beginning or the end @lamberta ?


lamberta:
For the [landing page](https://www.tensorflow.org/quantum ), rows need to be 3 or 4 cards wide ... with 4 looking best imho.

Can start a new row and move cards around to balance (and include an empty card), or start evicting old news. I think after 2-3 rows it starts looking funny.

MichaelBroughton:
Gotcha. Since we have six cards now, I put us at two rows of three cards.

lamberta (approved):
Looks right but may need to stage to make sure

PR #452, opened 2021-01-05
Adds support for resolving qubit ids in controlled gates

MichaelBroughton:
This is another step on the journey to closing #422 . Now we can resolve qubit ids to their integer values correctly in the `ResolveQubitIDs` functions in the controlled gate case. Next step is to implement changes in `circuit_parser_qsim.cc`.

I also cleaned up the tests so that they are much smaller and each one targets certain behavior more explicitly.


zaqqwerty (approved):
LGTM

PR #454, opened 2021-01-07
Noise tutorial

zaqqwerty:
I finished putting together a tutorial on how I am using TFQ to work with noise and density matrices.  Let me know how this looks, and if anything is not clear @QuantumVerd @MichaelBroughton.

Might we need to split the list of tutorials too?  (it is starting to get long)


MichaelBroughton (commented):
The notebook looks great, the tensor board visuals at the end are really nice!

Couple of thoughts:
1. There is a lot of math in this notebook and the code is pretty dense in a few places. For most of our tutorials we try to keep the amount of math and code density low to make them feel a bit more approachable (Our calculate gradients tutorial is best at this and our quantum_data tutorial is probably the worst at this). Would it be possible to cut down on the math and code density in places possibly at the cost of covering a little less ground ?

2. Before doing a more thorough review it might be worth cleaning up some of the language a bit to make sure it feels like a tutorial. A good example of this is using 2nd person (you instead of we). A few examples: 
> We can even take derivatives with respect to the mixture probabilities:
> Here we use the default tfq.differentiators.Adjoint for gradient backpropagation
> We start by calling tfq.layers.State to obtain the state

3. Again before doing a thorough review it might be worth cleaning up the use of excessive reshapings, dimension changes and use of obscure constants in the code:
```python3
with tf.GradientTape() as g:
  # Upgrade symbol_values to a tensor so that gradients can be tracked.
  symbol_values_t = tf.constant(symbol_values)
  g.watch(symbol_values_t)
  exp_list = tf.squeeze(tfq.layers.Expectation()(
    [c_a, c_b], symbol_names=[a, b],
    symbol_values=tf.tile(tf.expand_dims(symbol_values_t, 0), [2, 1]),
    operators=[[m], [m]]
  ))
  m_exp = tf.reduce_sum(tf.constant([p, 1-p]) * exp_list)
m_grad = g.gradient(m_exp, symbol_values_t)
print("Gradient of expectation value of M with respect to symbols a and b:")
print(m_grad)
```
Is this as simple as it can be ? Also: in some cases you can just write m_grad at the bottom and should only use print if you need to introduce formatting or when the `__repr__` and `__str__` methods are different.

Overall I think there's a lot of good material in here and it will take some amount of refinement to make tutorial ready. Just want to confirm a final time before we pour more work into the review here that we do indeed want to make this a tutorial (placing a great deal of value on understandability code simplicity, use of language, appropriate visuals etc. vs a research notebook where we intend these to be for more expert users) ?

zaqqwerty:
I've now added a new version of the tutorial.  I tried to address all comments so far, in particular,
- some math removed, for example the proof that every density matrix can be written as a probabilistic mixture of unitaries
- changed all the language to second person for a more tutorial style
- streamlined the code examples, for example no more usage of `tf.squeeze`
- changed all technical language according to Guillaume's suggestions

gatorwatt:
Another really helpful resource for visualizing error channels is offered in chapter 8 of Nielsen Chuang, section 8.3.3 which displays bloch sphere shape shifts from different channels. Perhaps a reference could be included? 

Also for the Tensorboard display at closing, I was kind of distracted by the smoothed curve in the scalars tab, not sure if there's any way to turn that off when initializing, also log plot looked prettier than the default.

MichaelBroughton:
Thanks for the suggestions @gatorwatt . In addition to those, now that we have the `tfq.noise` and noise incorporated into existing `tfq.layers` do we think it might make sense to revamp this tutorial a little bit to include some of these new features ?

zaqqwerty:
Superceded by #573

PR #455, opened 2021-01-07
Added multi qubit support to circuit_parser_qsim.

MichaelBroughton:
Almost done on #422 . This PR adds support for the `control_qubits` and `control_values` fields in the parsing layer for all of our gates. The last PR in this chain should be to just implement gates with `controlled_by` in our python tests and fix any last errors that pop up.


zaqqwerty (approved):
LGTM

PR #461, opened 2021-01-14
Expose batch deserializer

zaqqwerty:
Building some utilities that rely on converting tfq tensors to cirq, exposing what we use in tfq so I can use that


zaqqwerty:
For after #462

zaqqwerty:
@MichaelBroughton good point about the overlap between `_batch_deserialize_helper` and `tfq.from_tensor`, probably we can just use tfq.from_tensor

MichaelBroughton:
Going to close now since this PR has been resolved elsewhere. Feel free to re-open if something was missed.

PR #462, opened 2021-01-15
Extract custom sympy parsing

zaqqwerty:
Originally there was [a bug in Cirq](https://github.com/quantumlib/Cirq/issues/3192) which affected their `approx_eq` function.  However, the fix does not actually help our case - we actually want to know if the coefficients of symbolic multiplications are approximately equal.  I'm extracting the symbol parsing so it can be reused in tests elsewhere.


zaqqwerty:
Sorry for all the small pushes, I've had trouble with my local TensorFlow installation so I've been using the CI here for tests

MichaelBroughton (commented):
Took a look at: https://github.com/quantumlib/Cirq/issues/3192
Why can't the fix they have in there be used in our case ?

zaqqwerty:
```
q = cirq.GridQubit(0, 0)
s = sympy.Symbol("s")

no_symbol_1 = cirq.Circuit(cirq.rz(1.515)(q))
no_symbol_2 = cirq.Circuit(cirq.rz(1.510)(q))
print(cirq.approx_eq(no_symbol_1, no_symbol_2, atol=0.1))

symbol_1 = cirq.Circuit(cirq.rz(1.515*s)(q))
symbol_2 = cirq.Circuit(cirq.rz(1.510*s)(q))
cirq.approx_eq(symbol_1, symbol_2, atol=0.1)
```
The first approx_eq yields True; the second, instead of yielding False like before that issue was resolved, now yields `Insufficient information to decide whether expressions are approximately equal`, which is the correct behavior since the actual value of `s` is unknown (so no bound on the absolute tolerance exists).

The problem with either behavior is that in the test `test_resolve_parameters_consistency`, some different rounding of coefficients on symbols occurs in `tfq.resolve_parameters` versus `cirq.resolve_parameters`.  To compare them, we need to check that the coefficients are nearly the same instead of exactly the same.

tensorflow_quantum/python/util.py
```diff
@@ -344,6 +345,86 @@ def _symbols_in_op(op):
         "tfq.util.get_supported_gates().")
 
 
+def is_expression_approx_eq(exp_1, exp_2, atol):
```
MichaelBroughton, line 348: Can we turn this into `_is_expression_approx_eq` since it appears we only want people using `is_gate_approx_eq` ?

tensorflow_quantum/python/util.py
```diff
@@ -344,6 +345,86 @@ def _symbols_in_op(op):
...
+    """Compare possibly symbolic expressions for approximate equality.
+
+    Coefficient based approximate equality.  If no symbol is present in
```
MichaelBroughton, line 351: Might want to provide an args and return list since you are passing in some non trivial things.

tensorflow_quantum/python/util.py
```diff
@@ -344,6 +345,86 @@ def _symbols_in_op(op):
...
+    v_1 = serializer._scalar_extractor(exp_1)
+    v_2 = serializer._scalar_extractor(exp_2)
+    if isinstance(s_1, numbers.Real) and isinstance(s_2, numbers.Real):
+        s_eq = cirq.approx_eq(s_1, s_2, atol=atol)
+    elif isinstance(s_1, sympy.Symbol) and isinstance(s_2, sympy.Symbol):
+        s_eq = str(s_1) == str(s_2)
+    else:
+        s_eq = False
+    v_eq = cirq.approx_eq(v_1, v_2, atol=atol)
+    return s_eq and v_eq
```
MichaelBroughton, line 369:
```python3
v_eq = cirq.approx_eq(v1, v2, atol=atol)
if isinstance(s_1, numbers.Real) and isinstance(s_2, numbers.Real):
  return cirq.approx_eq(...) and v_eq
if isinstance(...) and isinstance(...):
  return str(s_1) == str(s_2) and v_eq
return False


```

tensorflow_quantum/python/util.py
```diff
@@ -344,6 +345,86 @@ def _symbols_in_op(op):
...
+
+
+def is_gate_approx_eq(gate_true, gate_deser, atol=1e-5):
```
MichaelBroughton, line 372: lets stick with something more conventional like cirq does with `cirq.approx_eq` and do `tfq.gate_approx_eq` to indicate the difference.

tensorflow_quantum/python/util.py
```diff
@@ -344,6 +345,86 @@ def _symbols_in_op(op):
...
+    """Compares gates in the allowed TFQ gate set.
+
+    Serialized gates can only have simple expressions of the form s*n or n*s,
+    where s is a `sympy.Symbol` and n is some number.  Checks that each
+    corresponding expression in `gate_true` and `gate_deser` are approximately
+    equal.
```
MichaelBroughton, line 378: Might want to rephrase here since serialization doesn't really impact the definition of our approximate equality. It might be simpler to just say something like: "Gates in TFQ support symbols, numbers or a single product of a real number and a symbol as their parameters. This function behaves like `cirq.approx_eq` specialized for these kinds of gates so that TFQ can support approximate equality in gates containing symbols: <snippet from your earlier comment in this PR> where as cirq does not." or something.

tensorflow_quantum/python/util.py
```diff
@@ -344,6 +345,86 @@ def _symbols_in_op(op):
...
+
+    This function is needed for testing round trips through the serializer,
+    since that trip rounds coefficients.
+
+    Args:
+        gate_true: `cirq.Gate` which is in the TFQ gate set.
+        gate_deser: `cirq.Gate` which is in the TFQ gate set.  This may be a
+            more general type of gate than `gate_true`, for example
+            `gate_true` may be an X while `gate_deser` is an XPowGate.
```
MichaelBroughton, line 387: We can probably just leave it as "`cirq.Gate` which is in the TFQ gate set.".

MichaelBroughton, line 387: Might actually want to mention that they work on gates found in tfq.util.get_supported_gates()

tensorflow_quantum/python/util.py
```diff
@@ -344,6 +345,86 @@ def _symbols_in_op(op):
...
+        bool which says if the two gates are approximately equal in the way
+            described above.
+    """
+    if not isinstance(gate_true, cirq.Gate) or not isinstance(
+            gate_deser, cirq.Gate):
+        raise TypeError("Arguments must be cirq gates.")
+    supported_gates = serializer.SERIALIZER.supported_gate_types()
+    if not (any([isinstance(gate_true, g) for g in supported_gates]) and
+            any([isinstance(gate_deser, g) for g in supported_gates])):
+        raise ValueError("One of the inputs is not a valid TFQ gate.")
```
MichaelBroughton, line 399: Does this work on ControlledGates like `cirq.X.controlled_by(....)` ?

tensorflow_quantum/python/util.py
```diff
@@ -344,6 +345,86 @@ def _symbols_in_op(op):
...
+            more general type of gate than `gate_true`, for example
+            `gate_true` may be an X while `gate_deser` is an XPowGate.
+
+    Returns:
+        bool which says if the two gates are approximately equal in the way
+            described above.
+    """
+    if not isinstance(gate_true, cirq.Gate) or not isinstance(
+            gate_deser, cirq.Gate):
+        raise TypeError("Arguments must be cirq gates.")
```
MichaelBroughton, line 395: Can we seperate these out so we can report to the user which argument wasn't a valid `cirq.Gate` in the case of an error ?

MichaelBroughton (changes requested):
Good start! After a little bit of cleanup and broadening out the tests to work on controlled gates we should be good to merge.

tensorflow_quantum/python/util.py
zaqqwerty: Done

tensorflow_quantum/python/util.py
zaqqwerty: Done

tensorflow_quantum/python/util.py
zaqqwerty: Done

tensorflow_quantum/python/util.py
zaqqwerty: Done

tensorflow_quantum/python/util.py
zaqqwerty: Done

tensorflow_quantum/python/util.py
zaqqwerty: Done

zaqqwerty: Added reference to tfq.util.get_supported_gates()

tensorflow_quantum/python/util.py
zaqqwerty: Separated the error messages

WORKSPACE
```diff
@@ -67,9 +67,9 @@ http_archive(
 
 http_archive(
     name = "qsim",
-    sha256 = "f390ee72cf88c48d81c98262c599dc45d660a2a9308a9ee903bfa73aec08a9b4",
-    strip_prefix = "qsim-0.6.0",
-    urls = ["https://github.com/quantumlib/qsim/archive/v0.6.0.zip"],
+    sha256 = "e2853379bde52d6277f9be4b80f54d32b3b27f7242a6c561cb34fb12d823b80e",
```
MichaelBroughton, line 70: There is some weird bleedover from old PRs working it's way into this one. And currently the CI tests aren't passing. I think this PR might've worked it's way into a pretty ugly state. Maybe it's worth starting a fresh branch off master and putting the changes you want in again ?

zaqqwerty:
Clean branch PR at #478

PR #469, opened 2021-02-04
made CZ ladder consistent with structure from McClean et al.

ryanhill1:
As it was, the CZ ladder was implemented as a simple chain, but this is not what they used in McClean et al. As you can see from McClean et al. Figure 2b. and the image you provided in the "Summary" section of the notebook, the CZ ladder is actually two layers of CZ  connections: the first connects odd-numbered qubits to the nearest neighbor even-numbered qubit, and the second connects even-numbered qubits to the nearest neighbor odd-numbered qubits. Thus, my edits make the CZ ladder structure consistent with that of McClean et al.

### DIFF
--- a/docs/tutorials/barren_plateaus.ipynb
+++ b/docs/tutorials/barren_plateaus.ipynb
@@ -1,501 +1,520 @@
 {
-  "cells": [
-    {
-      "cell_type": "markdown",
-      "metadata": {
-        "colab_type": "text",
-        "id": "xLOXFOT5Q40E"
-      },
-      "source": [
-        "##### Copyright 2020 The TensorFlow Authors."
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": 0,
-      "metadata": {
-        "cellView": "form",
-        "colab": {},
-        "colab_type": "code",
-        "id": "iiQkM5ZgQ8r2"
-      },
-      "outputs": [],
-      "source": [
-        "#@title Licensed under the Apache License, Version 2.0 (the \"License\");\n",
-        "# you may not use this file except in compliance with the License.\n",
-        "# You may obtain a copy of the License at\n",
-        "#\n",
-        "# https://www.apache.org/licenses/LICENSE-2.0\n",
-        "#\n",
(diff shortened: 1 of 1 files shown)


MichaelBroughton (commented):
Hi @ryanhill1 thanks for opening this PR. A Couple thoughts: 
1. The relative order of the CZ gates doesn't really matter in this case. That's because CZ gates only apply if the control qubit is a 1 and since the ladder just flips phases there's no chance that a qubit could go from being a 0 to 1 (it could just pickup some phase). I wrote myself a quick little check to verify this:
```python3
import cirq
import numpy as np

for n_qb in [4, 5, 6, 7, 8]:
    qubits = cirq.GridQubit.rect(1, n_qb)
    # Original approach
    c1 = cirq.Circuit(cirq.CZ(a, b) for a, b in zip(qubits, qubits[1:]))

    # Proposed approach.
    double = [qubits[i:i + 2] for i in range(0, len(qubits) - 1, 2)]
    c2 = cirq.Circuit()
    for i in range(len(double)):
        for src, dest in zip(double[i], double[i][1:]):
            c2 += cirq.CZ(src, dest)

    double_odd = [qubits[i:i + 2] for i in range(1, len(qubits) - 1, 2)]
    for i in range(len(double_odd)):
        for src, dest in zip(double_odd[i], double_odd[i][1:]):
            c2 += cirq.CZ(src, dest)

    np.testing.assert_array_equal(cirq.unitary(c1), cirq.unitary(c2))
```

2. The new approach is a fair bit more code in the tutorial to be more consistent with exactly what is shown in the figures of the paper. 

With all that being said as someone reading through the tutorial would you say you find it more intuitive to see an exact 1:1 matching with what happens in the figure at the cost of longer code (despite the order of the gates not really mattering based off of this intuition on CZ gates) ?

ryanhill1:
@MichaelBroughton, thanks for reviewing my PR so quickly. I did not realize the two CZ structures were equivalent, so thank you for demonstrating that in the example you gave. With that being case, it is definitely more intuitive to leave the tutorial as-is.

PR #470, opened 2021-02-04
Fix indentation in tutorial.

we-taper:
A small typo in the tutorial code.

It looks like the if statement should belong to the for loop.


MichaelBroughton (approved):
Good catch! The formatting on this tutorial does look a little weird in places... Hopefully no more issues like this pop up. Once you let the cla-bot know you've signed it we should be good to merge.

we-taper:
@googlebot I signed it!

PR #471, opened 2021-02-05
Removes use of old qubits

### LINKED ISSUE #467: Hello Many Worlds z0
This is a small thing and not positive it is in error, I was a little confused in hello many worlds tutorial seeing the z0 cirq object in the check_error(.) function definition, partly because it was defined in part 1 of the notebook and everything else in part 2 appeared to be defined from scratch. So yeah just thought would share.

MichaelBroughton:
In part 2 of the hello many worlds tutorial we used some paulisum operations that were defined on qubits defined in part 1. This changes things so that the paulisums defined in part 2 use qubits from part 2. Fixes #467


jaeyoo (approved):
LGTM

Issue #460, opened 2021-01-14
Upgrade to tf 2.4.0


MichaelBroughton:
Now that 2.4.0 (#456) is out it's time to upgrade! This will involve updating our various build scripts and bazel workspace files and fixing any bugs that come up.

PR #472, opened 2021-02-05
Upgrade nightly builds to tf 2.4.1.

MichaelBroughton:
Almost as painless as 2.3.0 -> 2.3.1 only instead of a micro version we jumped a minor version. This is the first half of addressing #460 . Interestingly I had to use 2.4.1 since 2.4.0 didn't appear to work after build+install. The second half will come in the form of an update to the install + docs when we cut an official release.

I've also modified the configure script since CPU and GPU are now all inside of the package `tensorflow` and I'm not 100% sure why we were making sure to distinguish between `tensorflow-cpu` and `tensorflow-gpu` @jaeyoo do you have any ideas ?

Note: Once this goes in we will need to update internal builds for nightly.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,6 +1,6 @@
 # This file includes external dependencies that are required to compile the
 # TensorFlow op. Maybe of them are specific versions used by the TensorFlow
-# binary used. These are extracted from TF v2.3.1, tensorflow/workspace.bzl.
+# binary used. These are extracted from TF v2.4.1, tensorflow/workspace.bzl.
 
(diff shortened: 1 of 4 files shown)


jaeyoo:
@MichaelBroughton The initial idea was the separation of TF custom op supports in cpu vs gpu versions. For example, tensorflow gpu automatically requires CUDA installed, and throw errors if they are not set. However, the issues may be resolved in the latest TF. let's figure it out with using this PR :)

jaeyoo (approved):
LGTM, I will work on Google internal changes after this one merged.

PR #473, opened 2021-02-06
Update install.md

jaeyoo:
Upgrade tf version to 2.4.1

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -138,10 +138,9 @@ Be sure the virtual environment you created in step 2 is activated. Then, instal
 
 <!-- common_typos_disable -->
 <pre class="devsite-click-to-copy">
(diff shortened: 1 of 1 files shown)


MichaelBroughton:
This looks good! Let's merge this in right before we cut a new 0.5.0 release. If we merge it now the instructions would get updated immediately which we don't want.

jaeyoo:
Sure! let's defer it to the new release day.

MichaelBroughton:
When we release, lets also remember to update the examples in the `research` branch to use 2.4.1. Too (a seperate PR, but not urgent at all).

MichaelBroughton:
In light of #482 maybe we should convert this PR to say that currently the latest install from pip is 2.3.1, but that the latest source build version is 2.4.1. What do you think ?

MichaelBroughton:
We've gone and cut the 0.5.0 release with updated install instructions so we can close this for now.

PR #474, opened 2021-02-06
Update inner_product_op.py

jaeyoo:
Fix the wrong test code & wrong latex expression in:

https://www.tensorflow.org/quantum/api_docs/python/tfq/math/inner_product?hl=en


MichaelBroughton (approved):
Thanks so much for the fix Jae and spotting the issues in our docs!

PR #477, opened 2021-02-08
Add adjoint differentiation of tfq.math.inner_product()

jaeyoo:
This PR adds adjoint differentiation of tfq.math_ops.inner_prod().

Since inner_product() has the specific output tensor shape `[batch_size, inner_size, n_symbols]`, this PR decided to implement new adjoint gradient op for it. (The original adjoint gradient has only `[batch_size, n_symbols]` output size, so one more internal nested for-loop is added in this new adjoint gradient op)

For edge cases, this PR deals with them like:

[1] empty symbols

The following code shows the default behavior of TensorFlow dealing with gradients with respect to the empty symbols.

```
import tensorflow as tf

>>> s = tf.constant([])
>>> with tf.GradientTape() as t:
>>>   t.watch(s)
>>>   x = tf.constant(tf.ones((3,4)))
>>> t.gradient(x, s)
None
```

So, this PR adds this behavior in Python `tf.custom_gradient()` function return function - `def grad(dy)`. For C++ `inner_product_adj_grad_op` itself, this PR adds throwing errors if the symbol is empty.

[2] empty circuits

I suspect that our current definition is controversial. for example, we can say that the given empty circuit is just `|0>`. Then, we can say that the output inner product is `1.0` for both empty circuits `<0|0>`, and its gradient is `0.0`. However, if only the `other_programs` is empty, we may have `<psi(x)|0>` and  `<dpsi(x)/dx|0>`.  That's why this PR didn't just return the default value when the circuit is empty.


tensorflow_quantum/core/ops/math_ops/inner_product_op.py
```diff
@@ -82,8 +161,60 @@ def inner_product(programs, symbol_names, symbol_values, other_programs):
...
+            return [None, None, None, None]
+        inner_prod_grad = inner_product_adj_grad(
+            programs, symbol_names, tf.cast(symbol_values, tf.float32),
+            other_programs)
+        return [None, None, tf.einsum("bos,bo->bos", inner_prod_grad, dy), None]
+
     return MATH_OP_MODULE.tfq_inner_product(programs, symbol_names,
                                             tf.cast(symbol_values, tf.float32),
-                                            other_programs)
+                                            other_programs), grad
```
MichaelBroughton, line 220: Normally we don't need to write the docstring + function for gradient functionality and we can just decorate the forward pass method with the `@RegisterGradient` decorator. See: https://www.tensorflow.org/api_docs/python/tf/RegisterGradient

tensorflow_quantum/core/ops/math_ops/tfq_inner_product_adj_grad.cc
```diff
@@ -0,0 +1,478 @@
...
+        for (std::vector<qsim::GateFused<QsimGate>>::size_type k = 0;
+             k < other_fused_circuits[i][j].size(); k++) {
+          qsim::ApplyFusedGate(sim, other_fused_circuits[i][j][k], scratch);
+        }
+
+        // now sv is |psi>, scratch is |phi>
+        // initialize gradients for given |psi> and |phi>.
+        for (int k = 0; k < maps[i].size(); k++) {
+          (*output_tensor)(i, j, k) = std::complex<float>(0, 0);
+        }
```
MichaelBroughton, line 236: Can just use `output_tensor.setZero()` before calling into ComputeSmall or ComputeLarge

tensorflow_quantum/core/ops/math_ops/tfq_inner_product_adj_grad.cc
```diff
@@ -0,0 +1,478 @@
...
+          // We've run into a new state vector we must compute.
+          // Only compute a new state vector when we have to.
+          if (nq > largest_nq) {
+            largest_nq = nq;
+            sv = ss.Create(largest_nq);
+            sv_adj = ss.Create(largest_nq);
+            scratch = ss.Create(largest_nq);
+            scratch2 = ss.Create(largest_nq);
+          }
+          // no need to update scratch_state since ComputeExpectation
```
MichaelBroughton, line 342: Is ComputeExpectation still here ?

tensorflow_quantum/core/ops/math_ops/tfq_inner_product_adj_grad.cc
```diff
@@ -0,0 +1,478 @@
...
+    for (const int num : num_qubits) {
+      max_num_qubits = std::max(max_num_qubits, num);
+    }
+
+    // Cross reference with standard google cloud compute instances
+    // Memory ~= 2 * num_threads * (2 * 64 * 2 ** num_qubits in circuits)
+    // e2s2 = 2 CPU, 8GB -> Can safely do 25 since Memory = 4GB
+    // e2s4 = 4 CPU, 16GB -> Can safely do 25 since Memory = 8GB
+    // ...
+    if (max_num_qubits >= 26 || output_dim_batch_size == 1) {
```
MichaelBroughton, line 168: Probably want to lower this to 24 since you are creating 4 state vectors inside.

tensorflow_quantum/core/ops/math_ops/tfq_inner_product.cc
```diff
@@ -182,12 +182,6 @@ class TfqInnerProductOp : public tensorflow::OpKernel {
         qsim::ApplyFusedGate(sim, fused_circuits[i][j], sv);
       }
       for (int j = 0; j < other_fused_circuits[i].size(); j++) {
-        // (#679) Just ignore empty program
-        if (fused_circuits[i].size() == 0) {
-          (*output_tensor)(i, j) = std::complex<float>(1, 0);
-          continue;
-        }
-
```
MichaelBroughton, line 190: Does removing this change the empty case behavior for the forward pass ?

MichaelBroughton (changes requested):
Hi Jae, this is amazing work!! Nice job adapting the adjoint method code to work on inner product.

I think there are some tweaks that need to be made to the code before we can merge:

1. The forward pass output shape is [batch, n_others], much like our expectation op which has forward pass: [batch, n_ops]. The gradient for our expectation op is the total gradient (i.e $sum_{op_i} [batch, n_ops, n_symbols] => [batch, n_symbols]$). In this case you aren't giving the total gradient summing over $other_i$ which would allow you to simplify the computation in the C++ op (you could have something like `AccumulateOperators`) that let's you accumulate all other_programs together so you can compute the total gradient for all states in one pass instead of one by one.

2. I know the way TensorFlow handles complex gradients can be a little complicated so I was wondering if you think it might make sense to setup a few tests with a small TF compute graph, using raw tf operations (no TFQ) to implement a small circuit inner product calculation and then compare the gradients that come out of that compute graph with the ones we produce from our op.

3. I think we may be able to remove a lot of the boilerplate function code for the gradient by just using the `@RegisterGradient` decorator.

tensorflow_quantum/core/ops/math_ops/tfq_inner_product.cc
jaeyoo: I thought that the configuration that the one program has cirq.X(q) and other_program is empty is possible (so, the answer should be 0.0 for <1|0>, not 1.0), but I got this error when I tried:

```
tensorflow.python.framework.errors_impl.InvalidArgumentError: A reference circuit contains qubits not found in paired circuit. [Op:TfqInnerProduct]
```

So, one is empty, the other should be empty. I missed it before. Please let me roll it back. thanks.

tensorflow_quantum/core/ops/math_ops/tfq_inner_product_adj_grad.cc
jaeyoo: Done.

tensorflow_quantum/core/ops/math_ops/tfq_inner_product_adj_grad.cc
jaeyoo: Oh, wrong comments. thank you for the finding. let me remove it.

tensorflow_quantum/core/ops/math_ops/tfq_inner_product_adj_grad.cc
jaeyoo: Done.

tensorflow_quantum/core/ops/math_ops/inner_product_op.py
jaeyoo: Let me try it...

jaeyoo: Hi Mike, I've tried `tf.RegisterGradient` instead of `tf.custom_gradient`, but it seems not supporting complex numbers.

```python
@tf.RegisterGradient("TfqInnerProduct")
def _tfq_inner_product_grad(op, grad):
    programs, symbol_names, symbol_values, other_programs = op.inputs
    if symbol_names.shape[0] == 0:
        return [None, None, None, None]
    inner_prod_grad = inner_product_adj_grad(
        programs, symbol_names, symbol_values, other_programs, grad)
    return [None, None, inner_prod_grad, None]
```

For example, the final result (x) was automatically converted into float, and lost their imaginary parts:
```
Mismatched elements: 28 / 30 (93.3%)
Max absolute difference: 1.05809069
Max relative difference: 0.99853275
 x: array([[ 4.114758e-03,  4.425383e-01, -7.175841e-02],
       [ 8.077559e-02,  1.215895e-02,  1.306026e-01],
       [ 5.921935e-01,  7.048310e-01,  1.323049e-09],...
 y: array([[ 4.109432e-03+7.442631e-03j,  4.425324e-01-7.890046e-01j,
        -7.175633e-02-2.537872e-01j],
       [ 8.077151e-02+4.973777e-01j,  1.216752e-02+2.243689e-01j,...
```

Some investigations said it's intended... [link1](https://github.com/tensorflow/tensorflow/issues/14081), [link2](https://github.com/tensorflow/tensorflow/issues/3346).

So I want to keep this `custom_gradient` implementation. WDYT?

MichaelBroughton: Ahhhh ok, thank you for checking. We can keep things using `custom_gradient`. I'll go ahead and review now.

MichaelBroughton (changes requested):
Things are looking pretty good now! A few more small tweaks and we should be ready to go. It will be exciting to start putting circuit inner product calculations inside of compute graphs and Keras models!

tensorflow_quantum/core/ops/math_ops/BUILD
```diff
@@ -82,3 +84,13 @@ py_test(
         "//tensorflow_quantum/python:util",
     ],
 )
+
+py_test(
+    name = "inner_product_adj_grad_op_test",
+    srcs = ["inner_product_adj_grad_op_test.py"],
```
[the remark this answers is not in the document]
jaeyoo: Done.

tensorflow_quantum/core/ops/math_ops/inner_product_adj_grad_op_test.py
```diff
@@ -0,0 +1,382 @@
...
+            [[resolver[symbol]
+              for symbol in symbol_names]
+             for resolver in resolver_batch])
+
+        programs = util.convert_to_tensor(circuit_batch)
+        other_programs = util.convert_to_tensor(other_batch)
+        symbol_names_tensor = tf.convert_to_tensor(symbol_names,
+                                                   dtype=tf.dtypes.string)
+        symbol_values = tf.convert_to_tensor(symbol_values_array)
+        prev_grad = np.random.normal(size=(batch_size, inner_dim_size)).astype(
```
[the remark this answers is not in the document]
jaeyoo: Done.

tensorflow_quantum/core/ops/math_ops/inner_product_adj_grad_op_test.py
```diff
@@ -0,0 +1,382 @@
...
+            'n_qubits': 10,
+            'batch_size': 10,  # ComputeSmall
+            'inner_dim_size': 2
+        },
+        {
+            'n_qubits': 5,
+            'batch_size': 10,  # ComputeSmall
+            'inner_dim_size': 5
+        },
+    ])
```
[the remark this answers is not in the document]
jaeyoo: Done.

tensorflow_quantum/core/ops/math_ops/inner_product_op.py
```diff
@@ -20,6 +20,83 @@
 MATH_OP_MODULE = load_module(os.path.join("math_ops", "_tfq_math_ops.so"))
 
 
+def inner_product_adj_grad(programs, symbol_names, symbol_values,
```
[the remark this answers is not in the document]
jaeyoo: Oh, i thought it was supposed to be exposed to users. Thank you for the catch. Done.

tensorflow_quantum/core/ops/math_ops/inner_product_op.py
```diff
@@ -61,7 +138,7 @@ def inner_product(programs, symbol_names, symbol_values, other_programs):
     Note: `other_programs` must not contain any free symbols. These can
         be resolved beforehand with `tfq.resolve_parameters`.
 
-    Note: Currently this op is not differentiable.
+    Note: Currently this op is differentiable via adjoint differentiation.
```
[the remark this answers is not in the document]
jaeyoo: Yes, I worried too much. Thank you for the confirmation. Done. :)

tensorflow_quantum/core/ops/math_ops/inner_product_op_test.py
```diff
@@ -312,6 +323,142 @@ def test_correctness_empty(self):
...
+        other_programs = util.convert_to_tensor(other_batch)
+        symbol_names = tf.convert_to_tensor([], dtype=tf.dtypes.string)
+        symbol_values = tf.convert_to_tensor([[] for _ in range(batch_size)])
+
+        with tf.GradientTape() as tape:
+            tape.watch(symbol_values)
+            ip = inner_product_op.inner_product(programs, symbol_names,
+                                                symbol_values, other_programs)
+        out = tape.gradient(ip, symbol_values)
+        self.assertIsNone(out)
```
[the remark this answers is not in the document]
jaeyoo: Oh great. Good to know. Thanks! Done.

tensorflow_quantum/core/ops/math_ops/inner_product_op.py
```diff
@@ -82,8 +159,15 @@ def inner_product(programs, symbol_names, symbol_values, other_programs):
...
             resolved in and `other_programs[i][j]`.
-
     """
+
+    def grad(dy):
+        if symbol_names.shape[0] == 0:
+            return [None, None, None, None]
+        inner_prod_grad = inner_product_adj_grad(
+            programs, symbol_names, symbol_values, other_programs, dy)
+        return [None, None, inner_prod_grad, None]
```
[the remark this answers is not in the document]
jaeyoo: Added `tf.cond()`. Done.

tensorflow_quantum/core/ops/math_ops/inner_product_adj_grad_op_test.py
```diff
@@ -0,0 +1,382 @@
...
+        prev_grad = np.ones((batch_size, inner_dim_size))
+
+        with self.assertRaisesRegex(tf.errors.InvalidArgumentError,
+                                    'symbols must be a positive integer'):
+            inner_product_op.inner_product_adj_grad(programs, symbol_names,
+                                                    symbol_values,
+                                                    other_programs,
+                                                    prev_grad)
+
+    def test_correctness_empty(self):
```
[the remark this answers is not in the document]
jaeyoo: Added. Done.

jaeyoo:
PTAL.

MichaelBroughton (approved):
LGTM

tensorflow_quantum/core/ops/math_ops/inner_product_op.py
```diff
@@ -82,8 +129,19 @@ def inner_product(programs, symbol_names, symbol_values, other_programs):
...
+
+    def grad(dy):
+
+        def _true_grad():
+            return _inner_product_adj_grad(programs, symbol_names,
+                                           symbol_values, other_programs, dy)
+
+        inner_prod_grad = tf.cond(tf.math.equal(symbol_names.shape[0], 0),
+                                  lambda: tf.zeros_like(symbol_values),
+                                  _true_grad)
```
MichaelBroughton, line 142: I don't think `symbol_names.shape[0]` is accesible in non-eager mode. Would you be able to change this so that it looks more like: https://github.com/tensorflow/quantum/blob/master/tensorflow_quantum/python/differentiators/differentiator.py#L36 ?
Where we use `tf.size` and `tf.equal` instead ?

tensorflow_quantum/core/ops/math_ops/inner_product_grad_test.py
```diff
@@ -0,0 +1,385 @@
...
+        empty_symbols = tf.raw_ops.Empty(shape=(0,), dtype=tf.string)
+        empty_values = tf.raw_ops.Empty(shape=(0, 0), dtype=tf.float32)
+        other_program = tf.raw_ops.Empty(shape=(0, 0), dtype=tf.string)
+        empty_pred_grad = tf.raw_ops.Empty(shape=(0, 0), dtype=tf.float32)
+
+        with self.assertRaisesRegex(tf.errors.InvalidArgumentError,
+                                    'number of symbols must be a positive'):
+            _ = inner_product_op._inner_product_adj_grad(
+                empty_circuit, empty_symbols, empty_values, other_program,
+                empty_pred_grad)
```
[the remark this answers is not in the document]
jaeyoo: Done.

tensorflow_quantum/core/ops/math_ops/inner_product_op.py
jaeyoo: Done.

MichaelBroughton (approved):
LGTM

PR #478, opened 2021-02-08
Custom gate_approx_eq

zaqqwerty:
Add functions to approximately compare parameterized gates in the TFQ gate set.

Supersedes #462 .  In this PR, full support for ControlledGates is added.


MichaelBroughton (changes requested):
Looks good. Only one small comment is could you separate out the error testing for each function into separate tests ? I know we aren't great about doing this everywhere in the codebase, but smaller tests are generally easier to add onto or tweak etc.

MichaelBroughton (approved):
LGTM

PR #483, opened 2021-02-19
Empty support

MichaelBroughton:
Looks like we missed the case of empty tensors (we were very good about empty circuits) in some key places in the library. This PR adds in explicit empty tensor tests for all ops and differentiators.

### DIFF
--- a/tensorflow_quantum/core/ops/math_ops/inner_product_op_test.py
+++ b/tensorflow_quantum/core/ops/math_ops/inner_product_op_test.py
@@ -315,7 +315,6 @@ class InnerProductTest(tf.test.TestCase, parameterized.TestCase):
     def test_correctness_no_circuit(self):
         """Test the inner product between no circuits."""
 
-        # empty_cicuit = util.convert_to_tensor([cirq.Circuit()])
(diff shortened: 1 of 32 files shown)


tensorflow_quantum/core/ops/math_ops/inner_product_op_test.py
```diff
@@ -312,6 +312,19 @@ def test_correctness_empty(self):
         expected = np.array([[1.0]], dtype=np.complex64)
         self.assertAllClose(out, expected)
 
+    def test_correctness_no_circuit(self):
+        """Test the inner product between no circuits."""
+
+        # empty_cicuit = util.convert_to_tensor([cirq.Circuit()])
```
jaeyoo, line 318: nit: remove comment

jaeyoo (approved):
LGTM

tensorflow_quantum/core/ops/math_ops/inner_product_op_test.py
MichaelBroughton: Fixed.

jaeyoo (approved):

PR #491, opened 2021-02-24
Add nightly consistency checks for Cirq

MichaelBroughton:
Now we test against Cirq master and should be able to flag breakages as they happen.
Related: https://github.com/quantumlib/Cirq/issues/3828


balopat (dismissed):
LGTM, thank you for adding this!

PR #495, opened 2021-03-02
Update TFQ Python utils with input validation and docstring consistency

amogh7joshi:
Updated `tensorflow_quantum.python.utils.py` with improved docstring consistency (e.g., kept language format consistent, updated a couple of typos, and added information about error raising), and input validation for some methods.

### DIFF
--- a/tensorflow_quantum/python/util.py
+++ b/tensorflow_quantum/python/util.py
@@ -489,7 +489,6 @@ def gate_approx_eq(gate_true, gate_deser, atol=1e-5):
         TypeError: If input gates are not of type `cirq.Gate`.
         ValueError: If invalid gate types are provided.
     """
-
     if not isinstance(gate_true, cirq.Gate):
         raise TypeError(f"`gate_true` not a cirq gate, got {type(gate_true)}")
     if not isinstance(gate_deser, cirq.Gate):
@@ -544,7 +543,12 @@ def get_circuit_symbols(circuit):
 
     Returns:
         Python `list` containing the symbols found in the circuit.
+    
+    Raises:
+        TypeError: If `circuit` is not of type `cirq.Circuit`.
     """
+    if not isinstance(circuit, cirq.Circuit):
+        raise TypeError(f"Expected a cirq.Circuit object, got {circuit}.")
     all_symbols = set()
     for moment in circuit:
         for op in moment:
--- a/tensorflow_quantum/python/util.py
+++ b/tensorflow_quantum/python/util.py
@@ -256,7 +256,7 @@ def convert_to_tensor(items_to_convert):
             or `cirq.Paulisum` objects. Must be recangular.
     Returns:
         A `tf.Tensor` that represents the input items.
-        
+
     Raises:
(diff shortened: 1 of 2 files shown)


tensorflow_quantum/python/util.py
```diff
@@ -529,7 +543,12 @@ def get_circuit_symbols(circuit):
 
     Returns:
         Python `list` containing the symbols found in the circuit.
+
+    Raises:
+        TypeError: If `circuit` is not of type `cirq.Circuit`.
     """
+    if not isinstance(circuit, cirq.Circuit):
+        raise TypeError(f"Expected a cirq.Circuit object, got {circuit}.")
```
MichaelBroughton, line 551: Would you be able to add a test to this new behavior ?

MichaelBroughton (changes requested):
Thanks so much for the cleanup. Always appreciated!

Would you be able to add a test in for the new error mechanism you added in the `get_circuit_symbols` ? Once we've got that, should be good to merge.

amogh7joshi:
I added the test case, so let me know if it will work, or if anything else is necessary.

amogh7joshi:
@MichaelBroughton, the Formatting check seems to be continuously failing without much detail as to the actual issue. Could you let me know what there actually is to fix here?

MichaelBroughton:
Hi @amogh7joshi you can go over to the `checks` tab of this PR and look at the outputs for the formatting test for your latest version of the code:
image

Additionally when you're making these changes you can also just run `./scripts/format_check.sh` and `./scripts/lint_all.sh` to check your local changes are formatted and linted correctly.

Hope this helps.

amogh7joshi:
@MichaelBroughton, I ran the lint and formatting checks locally, and they completed without errors. Also, the only actual difference between the "expected" and "actual" formats is that there is a trailing whitespace after one of the lines. I added the whitespace, but this then leads to the lint check erroring out. I'm not quite sure what to do from here.

MichaelBroughton:
Ok. So I checked out your branch `patch-1` and found the following things:
1. On line 410 of `util_test.py` that reads: `expected_regex='Expected a ' ` you have an extra space at the end of the line. This needs to be removed (we don't want lingering unnecessary white space in files if it can be avoided so it's actually correctly reporting this).
2. On line 411 of `util_test.py` you have a single line break before the beginning of the new class `ExponentialUtilFunctionsTest`. Add another space (as per the [style guide](https://google.github.io/styleguide/pyguide.html)).

Locally you should have the same version of pylint and yapf that are specified in the [requirements.txt](https://github.com/tensorflow/quantum/blob/master/requirements.txt) file of our repo. When calling `format_check` and `lint_all` I found no errors reported, so if after making these changes you are still getting these errors you might want to double check your installed versions.

amogh7joshi:
Ah, I see the issues that you're pointing out. I think I was confusing the formatting check, thinking that the errored output was the correct output and the correct output was the errored output, so I was doing the opposite of what I was supposed to be doing. Thanks for letting me know, I've fixed it now and it should work properly.

MichaelBroughton (approved):
LGTM

PR #496, opened 2021-03-04
Upgrade qsim to 0.8.0.

MichaelBroughton:
Pretty straightforward upgrade, should enable #487 to merge in. Won't be our last qsim upgrade in the next little while since we do still need support on https://github.com/quantumlib/qsim/issues/288 in a release.

Note: qsim swapped indexing for unitaryspace so that's why we make the index change.


zaqqwerty (approved):
LGTM

PR #498, opened 2021-03-06
Bump cirq to 0.10.0.

MichaelBroughton:


release/setup.py
```diff
@@ -50,7 +50,9 @@ def finalize_options(self):
             self.install_lib = self.install_platlib
 
 
-REQUIRED_PACKAGES = ['cirq == 0.9.1', 'sympy == 1.5']
+REQUIRED_PACKAGES = [
```
jaeyoo, line 53: I realized that this can be required from the empty virtualenv:

```
google-api-core==1.21.0
google-auth==1.18.0
googleapis-common-protos==1.52.0
grpcio==1.30.0
protobuf==3.12.0
```

jaeyoo (approved):
LGTM with minor suggestion.

release/setup.py
jaeyoo: I tried to install `cirq==0.10.1` last week and failed, and found out this workaround.

MichaelBroughton: All these dependencies worked. Although I did need to use protobuf version 3.13.0 instead of 3.12.0 to have everything work.

jaeyoo: Sounds great. LGTM.

jaeyoo (approved):

verult:
Is it possible to require minimum versions instead? If someone uses TFQ in conjunction with another library that requires a higher version than what's included here, the other library might break.

In the TFQ multi-node tutorial, the latest `google-cloud-storage` package requires `google-auth >= 1.24.0`, and a missing package showed up during execution.

PR #501, opened 2021-03-08
[Noise] Adds noisy versions of util circuit functions.

MichaelBroughton:
Adds the ability for `random_***_circuit` and batch related methods to return circuits that contain channels.
Enables noisy tests on util ops.

### DIFF
--- a/tensorflow_quantum/python/util.py
+++ b/tensorflow_quantum/python/util.py
@@ -107,26 +107,33 @@ def _apply_random_control(gate, all_qubits):
 
 def random_symbol_circuit(qubits,
                           symbols,
+                          *,
                           n_moments=15,
                           p=0.9,
-                          include_scalars=True):
+                          include_scalars=True,
+                          include_channels=False):
     """Generates a random circuit including some parameterized gates.
 
     Symbols are randomly included in the gates of the first `n_moments` moments
     of the resulting circuit.  Then, parameterized H gates are added as
     subsequent moments for any remaining unused symbols.
     """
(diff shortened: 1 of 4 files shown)


tensorflow_quantum/python/util.py
```diff
@@ -467,6 +489,16 @@ def _expression_approx_eq(exp_1, exp_2, atol):
...
 
 
+# TODO: replace with cirq.approx_eq once
+# https://github.com/quantumlib/Cirq/issues/3886 is resolved for all channels.
+def _channel_approx_eq(op_true, op_deser, atol=1e-5):
+    if isinstance(op_true, cirq.DepolarizingChannel):
+        if isinstance(op_deser, cirq.DepolarizingChannel):
+            return abs(op_true.p - op_deser.p) < atol
+
+    return False
```
jaeyoo, line 499: nit:`raise TypeError("op_true and op_deser should be of cirq.DepolorizingChannel"`, to distinguish the right False and wrong type cases.

jaeyoo (approved):
LGTM with nits.

tensorflow_quantum/python/util.py
MichaelBroughton: In this case if a Channel is supplied that isn't supported, you will still raise the `raise ValueError(
        f"Some valid TFQ gate type is not yet accounted for, got {gate_true}")` at the end. Right now there is no support for other channels so it will actually behave correctly if an unknown channel is supplied.

PR #502, opened 2021-03-10
[Noise] Added NoisyQsimCircuit parsing in C++.

MichaelBroughton:
Adds `NoisyQsimCircuitFromProto` function for generating the new `qsim::NoiseyCircuit<Channel>` type circuits. This method notably does not have a fuser or metadata tracking (for adjoint differenitation) built in since the API that qsim currently [exposes for trajectory simulation](https://github.com/quantumlib/qsim/blob/master/lib/qtrajectory.h) is somewhat limited (no ability to manually fuse ahead of time or uncompute from a given trajectory back to zero for arbitrary channels, although mixtures are supported).

Also upgrades us to qsim 0.9.1 to handle: https://github.com/quantumlib/qsim/pull/308


jaeyoo (approved):
LGTM

PR #503, opened 2021-03-10
Add symbol_names empty check to diffs

MichaelBroughton:
Adds an empty check to the `tf.cond` that checks if the `symbol_names` tensor is empty to avoid accidentally computing lots of circuit info, only to not use any of it.


jaeyoo (approved):
LGTM

PR #505, opened 2021-03-16
Tutorial mnist: improve 1.3 Remove contradictory

Bankde:
The current code gives incorrect data count on "Number of 3s" and "Number of 6s"
```
Number of unique images: 10387
Number of 3s:  49
Number of 6s:  49
Number of contradictory images:  49

Initial number of examples:  12049
Remaining non-contradictory examples:  10338
```

This is because `new_y.append(labels.pop())` changes the mapping while the counting is also based on mapping.
`num_3 = sum(1 for value in mapping.values() if True in value)`

#### Why not just reverting back to `new_y.append(list(labels)[0])`?
Ans: Because there are (x,y) that are duplicated (same x and y). Unlike `list(labels)[0]`, `labels.pop()` effectively removes those duplicates so I try to preserve this behavior by also iterating mapping instead of (x,y).

I also slipped in the `next(iter(labels))` because it generally runs faster than converting to list[0]. I could revert it back if you suggest that list[0] gives better readability though.

New code: (10338 = (4961-49) + (5475-49))
```
Number of unique images: 10387
Number of 3s:  4961
Number of 6s:  5475
Number of contradictory images:  49

Initial number of examples:  12049
Remaining non-contradictory examples:  10338
```


MichaelBroughton (changes requested):
Hi @Bankde thanks for spotting this. After taking a closer look at the notebook I think this might be an opportunity to improve things even more! When I read:

```
Number of unique images: 10387
Number of 3s:  4961
Number of 6s:  5475
Number of contradictory images:  49

Initial number of examples:  12049
Remaining non-contradictory examples:  10338
```
I think OK, there's 10387 unique images in this Dataset. 4961 threes 5475 sixes and 49 duplicates. Adding those up gives me: 10485 (off by 49) which might take a moment for the reader to figure out that the 49 from below is actually baked into the number of 3s and 6s. Further to that when I read the next section I get "Remaining non-contradictory examples: 10388" is also a little confusing since I don't know what quantities from above that includes. What if instead we changed the reporting to something like:

```
Total number of unique images: 10387
Number of unique 3s:  4912
Number of unique 6s:  5426
Number of unique contradicting labels (both 3 and 6):  49

Initial number of images:  12049
Remaining non-contradicting unique examples:  10338
```
In here I see that `4912 + 5426 + 49 == 10387` and that  "Remaining non-contradicting unique examples" means I should add up the number of 3s and 6s (`10338 == 4912 + 5426`) which feels like it's a lot clearer for our readers. What do you think ?

Bankde:
Hi, @MichaelBroughton. Thanks for reviewing. I agree it sounds better so I have pushed the improvement.

MichaelBroughton (approved):
LGTM

PR #511, opened 2021-03-24
Add internal change

jaeyoo:
This is required for google internal changes.

### DIFF
--- a/tensorflow_quantum/core/ops/noise/BUILD
+++ b/tensorflow_quantum/core/ops/noise/BUILD
@@ -58,11 +58,15 @@ cc_binary(
     }),
     linkshared = 1,
     deps = [
+        # cirq cc proto
         "//tensorflow_quantum/core/ops:parse_context",
         "//tensorflow_quantum/core/ops:tfq_simulate_utils",
         "//tensorflow_quantum/core/src:circuit_parser_qsim",
         "//tensorflow_quantum/core/src:util_qsim",
         "@qsim//lib:qsim_lib",
+        # tensorflow core framework
+        # tensorflow core lib
+        # tensorflow core protos
     ],
 )
 
--- a/tensorflow_quantum/core/ops/noise/BUILD
+++ b/tensorflow_quantum/core/ops/noise/BUILD
@@ -1,3 +1,5 @@
+# load op_wrapper
+
 package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"])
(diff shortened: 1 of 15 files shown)


tensorflow_quantum/core/proto/BUILD
```diff
@@ -1,11 +1,9 @@
+load("@com_google_protobuf//:protobuf.bzl", "py_proto_library")
+
 package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"])
 
-load("@com_google_protobuf//:protobuf.bzl", "py_proto_library")
-
-# load("//tools/build_defs/proto/cpp:cc_proto_library.bzl", "cc_proto_library")
```
MichaelBroughton, line 7: Should we remove this ?

tensorflow_quantum/core/ops/noise/BUILD
```diff
@@ -56,11 +58,15 @@ cc_binary(
...
+        # cirq cc proto
         "//tensorflow_quantum/core/ops:parse_context",
         "//tensorflow_quantum/core/ops:tfq_simulate_utils",
-        "//tensorflow_quantum/core/src:util_qsim",
         "//tensorflow_quantum/core/src:circuit_parser_qsim",
+        "//tensorflow_quantum/core/src:util_qsim",
         "@qsim//lib:qsim_lib",
+        # tensorflow core framework
+        # tensorflow core lib
+        # tensorflow core protos
```
MichaelBroughton, line 69: Do we need the comments ?

MichaelBroughton (commented):
Looks good, just wondering if we should be removing the `py_proto_library` rule. There's also a few comments laying around in the files that should probably be removed.

tensorflow_quantum/core/ops/noise/BUILD
jaeyoo: Yes, it's for internal ones.

tensorflow_quantum/core/proto/BUILD
jaeyoo: Oops, nice catch. it should be removed.

jaeyoo: I mean, the comment should be removed because it's internal. Also, the public rule at the line 5 should be moved to the line 1 due to internal BUILD file linter.

tensorflow_quantum/core/ops/noise/BUILD
jaeyoo: It's syntactic sugar for copybara.

MichaelBroughton (approved):

PR #518, opened 2021-03-25
Adds tfq.noise.samples op.

MichaelBroughton:
Like with our noisy expectation op, we now have a noisy sampling op. The parallelization for this op was a little bit simpler than with the expectation op.

### DIFF
--- a/tensorflow_quantum/core/ops/noise/noisy_samples_op_test.py
+++ b/tensorflow_quantum/core/ops/noise/noisy_samples_op_test.py
@@ -56,85 +56,80 @@ class NoisySamplingTest(tf.test.TestCase, parameterized.TestCase):
         with self.assertRaisesRegex(tf.errors.InvalidArgumentError,
                                     'rank 1. Got rank 2'):
             # programs tensor has the wrong shape.
-            noisy_samples_op.samples(
-                util.convert_to_tensor([circuit_batch]), symbol_names,
-                symbol_values_array, [num_samples])
+            noisy_samples_op.samples(util.convert_to_tensor([circuit_batch]),
+                                     symbol_names, symbol_values_array,
+                                     [num_samples])
 
         with self.assertRaisesRegex(tf.errors.InvalidArgumentError,
                                     'rank 1. Got rank 2'):
             # symbol_names tensor has the wrong shape.
-            noisy_samples_op.samples(
-                util.convert_to_tensor(circuit_batch), np.array([symbol_names]),
-                symbol_values_array, [num_samples])
+            noisy_samples_op.samples(util.convert_to_tensor(circuit_batch),
+                                     np.array([symbol_names]),
+                                     symbol_values_array, [num_samples])
 
         with self.assertRaisesRegex(tf.errors.InvalidArgumentError,
                                     'rank 2. Got rank 3'):
             # symbol_values tensor has the wrong shape.
-            noisy_samples_op.samples(
-                util.convert_to_tensor(circuit_batch), symbol_names,
-                np.array([symbol_values_array]), [num_samples])
+            noisy_samples_op.samples(util.convert_to_tensor(circuit_batch),
+                                     symbol_names,
+                                     np.array([symbol_values_array]),
+                                     [num_samples])
 
         with self.assertRaisesRegex(tf.errors.InvalidArgumentError,
                                     'rank 2. Got rank 1'):
             # symbol_values tensor has the wrong shape 2.
-            noisy_samples_op.samples(
-                util.convert_to_tensor(circuit_batch), symbol_names,
-                symbol_values_array[0], [num_samples])
+            noisy_samples_op.samples(util.convert_to_tensor(circuit_batch),
+                                     symbol_names, symbol_values_array[0],
+                                     [num_samples])
 
         with self.assertRaisesRegex(tf.errors.InvalidArgumentError,
                                     'rank 1. Got rank 2'):
             # num_samples tensor has the wrong shape.
-            noisy_samples_op.samples(
-                util.convert_to_tensor(circuit_batch), symbol_names,
-                symbol_values_array, [[num_samples]])
+            noisy_samples_op.samples(util.convert_to_tensor(circuit_batch),
+                                     symbol_names, symbol_values_array,
+                                     [[num_samples]])
 
         with self.assertRaisesRegex(tf.errors.InvalidArgumentError,
                                     'Unparseable proto'):
             # programs tensor has the right type, but invalid value.
-            noisy_samples_op.samples(['junk'] * batch_size,
-                                                  symbol_names,
-                                                  symbol_values_array,
-                                                  [num_samples])
+            noisy_samples_op.samples(['junk'] * batch_size, symbol_names,
+                                     symbol_values_array, [num_samples])
 
         with self.assertRaisesRegex(tf.errors.InvalidArgumentError,
                                     'Could not find symbol in parameter map'):
             # symbol_names tensor has the right type, but invalid value.
-            noisy_samples_op.samples(
-                util.convert_to_tensor(circuit_batch), ['junk'],
-                symbol_values_array, [num_samples])
+            noisy_samples_op.samples(util.convert_to_tensor(circuit_batch),
+                                     ['junk'], symbol_values_array,
+                                     [num_samples])
 
         with self.assertRaisesRegex(TypeError, 'Cannot convert'):
             # programs tensor has the wrong type.
-            noisy_samples_op.samples([1] * batch_size,
-                                                  symbol_names,
-                                                  symbol_values_array,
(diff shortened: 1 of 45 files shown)


tensorflow_quantum/core/ops/noise/noisy_samples_op_test.py
```diff
@@ -0,0 +1,289 @@
...
+
+        with self.assertRaisesRegex(tf.errors.UnimplementedError,
+                                    'Cast string to float is not supported'):
+            # programs tensor has the wrong type.
+            noisy_samples_op.samples(util.convert_to_tensor(circuit_batch),
+                                     symbol_names, [['junk']] * batch_size,
+                                     [num_samples])
+
+        with self.assertRaisesRegex(Exception, 'junk'):
+            # num_samples tensor has the wrong shape.
```
jaeyoo, line 123: wrong type

tensorflow_quantum/core/ops/noise/tfq_noisy_samples.cc
```diff
@@ -0,0 +1,348 @@
...
+    auto construct_f = [&](int start, int end) {
+      for (int i = start; i < end; i++) {
+        auto r = NoisyQsimCircuitFromProgram(
+            programs[i], maps[i], num_qubits[i], true, &qsim_circuits[i]);
+        if (!r.ok()) {
+          p_lock.lock();
+          parse_status = r;
+          p_lock.unlock();
+          return;
+        }
```
jaeyoo, line 93: `NESTED_FN_STATUS_SYNC` after merging https://github.com/tensorflow/quantum/pull/519

tensorflow_quantum/core/ops/noise/tfq_noisy_samples.cc
```diff
@@ -0,0 +1,348 @@
...
+    // samples written by thread range [0, i].
+    std::vector<std::vector<long>> offset_prefix_sum(
+        num_threads, std::vector<long>(output_dim_batch_size, 0));
+
+    for (int i = 0; i < output_dim_batch_size; i++) {
+      int p_reps = (num_samples + num_threads - 1) / num_threads;
+      offset_prefix_sum[0][i] = rep_offsets[0][i] + p_reps;
+      for (int j = 1; j < num_threads; j++) {
+        offset_prefix_sum[j][i] += offset_prefix_sum[j - 1][i];
+        offset_prefix_sum[j][i] += rep_offsets[j][i] + p_reps;
```
jaeyoo, line 237: I don't understand why `offset_prefix_sum[j][i]` varies over `batch_size`. Since we have the same `num_samples` for all batch, can we just use `offset_prefix_sum[j]`?

tensorflow_quantum/core/ops/noise/noisy_samples_op_test.py
```diff
@@ -0,0 +1,289 @@
...
+        qubits1 = cirq.GridQubit.rect(1, n_qubits)
+        qubits2 = cirq.GridQubit.rect(1, n_qubits + 1)
+
+        circuit_batch1, resolver_batch = \
+            util.random_circuit_resolver_batch(
+                qubits1, batch_size, include_channels=True)
+
+        circuit_batch2, _ = \
+            util.random_circuit_resolver_batch(
+                qubits2, batch_size, include_channels=True)
```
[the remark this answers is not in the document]
MichaelBroughton: Fixed.

tensorflow_quantum/core/ops/noise/noisy_samples_op_test.py
```diff
@@ -0,0 +1,289 @@
...
+        b_reps = np.asarray(op_samples[2:])
+        self.assertEqual(a_reps.shape, (2, 10, 5))
+        self.assertEqual(b_reps.shape, (2, 10, 6))
+
+    def test_correctness_empty(self):
+        """Test the expectation for empty circuits."""
+        empty_circuit = util.convert_to_tensor([cirq.Circuit()])
+        empty_symbols = tf.convert_to_tensor([], dtype=tf.dtypes.string)
+        empty_values = tf.convert_to_tensor([[]])
+        empty_n_samples = tf.convert_to_tensor([1], dtype=tf.int32)
```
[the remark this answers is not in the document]
MichaelBroughton: The `num_samples` parameter doesn't have a `batch_size` first dimension so regardless of batch_size this will always need to have size 1 for valid inputs. So in this input batch_size is zero meaning empty, but we do need to keep one entry in this tensor.

tensorflow_quantum/core/ops/noise/noisy_samples_op_test.py
```diff
@@ -0,0 +1,289 @@
...
+
+        expected = np.array([[[]]], dtype=np.int8)
+        self.assertAllClose(out.to_tensor(), expected)
+
+    def test_correctness_no_circuit(self):
+        """Test the correctness with the empty tensor."""
+        empty_circuit = tf.raw_ops.Empty(shape=(0,), dtype=tf.string)
+        empty_symbols = tf.raw_ops.Empty(shape=(0,), dtype=tf.string)
+        empty_values = tf.raw_ops.Empty(shape=(0, 0), dtype=tf.float32)
+        empty_n_samples = tf.convert_to_tensor([1], dtype=tf.int32)
```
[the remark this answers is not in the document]
MichaelBroughton: See above :)

tensorflow_quantum/core/src/util_qsim_test.cc
```diff
@@ -714,5 +714,36 @@ TEST(UtilQsimTest, BalanceTrajectoryFewHigh) {
...
+  std::vector<std::vector<int>> tmp(offsets[0].size(),
+                                    std::vector<int>(2, n_reps));
+  BalanceTrajectory(n_reps, num_threads, &offsets);
+  AssertWellBalanced(tmp, num_threads, offsets);
+}
+
+TEST(UtilQsimTest, BalanceTrajectory1D_2) {
+  const int n_reps = 11;
+  const int num_threads = 10;
+  // [num_threads, n_reps.size()]
```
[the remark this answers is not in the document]
MichaelBroughton: Sorry, the second dimension in the 1d case is just the `batch_size` and can be arbitrary.

tensorflow_quantum/core/src/util_qsim_test.cc
```diff
@@ -714,5 +714,36 @@ TEST(UtilQsimTest, BalanceTrajectoryFewHigh) {
   AssertWellBalanced(n_reps, num_threads, offsets);
 }
 
+TEST(UtilQsimTest, BalanceTrajectory1D) {
+  const int n_reps = 100;
+  const int num_threads = 5;
+  // [num_threads, n_reps.size()]
```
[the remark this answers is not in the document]
MichaelBroughton: See above.

tensorflow_quantum/core/ops/noise/noisy_samples_op_test.py
MichaelBroughton: Fixed.

tensorflow_quantum/core/ops/noise/tfq_noisy_samples.cc
MichaelBroughton: This is a little tricky. `offset_pref_sum[j][i]` stores the `final sample number i that thread j should draw` If I do `offset_pref_sum[j][i] - offset_prefix_sum[j-1][i]` I get `the number of samples that thread j should draw from circuit i`. And finally `offset_prefix_sum[j-1][i]` tells `thread j where to start storing sample i` that it draws from the circuit. This allows one to do non locking writes to the output tensor since each thread knows exactly which indices in the output tensor it needs to write to.

`BalanceTrajectory` divides up the work across the threads evenly so in the event of: `n_samples = 10`, `n_circuits = 9` and `num_threads=11` The BalanceTrajectory function will assign offsets so that threads [1,10] all draw 1 sample from the first circuit with the 11th thread drawing no samples from the first circuit. To balance this out the 11th thread would then draw two samples from the second circuit and so on.... so the values of offset_prefix_sum[j][i] might be very close to one another, but they do have small offsets to make sure things are balanced and we can't just use `offset_prefix_sum[j]`. 

`offset_prefix_sum` stores all the information for the threads to do this balancing between them.

Does this clear things up ?

jaeyoo: Two thumbs up!

jaeyoo (approved):
LGTM

tensorflow_quantum/core/ops/noise/tfq_noisy_samples.cc
jaeyoo: Since it's merged, please apply this at https://github.com/tensorflow/quantum/pull/519

PR #519, opened 2021-03-25
Improved nested OP_REQUIRES_OK macros.

MichaelBroughton:
The `OP_REQUIRES_OK` macros have been flaking on ocassion when used in functions that get called in parallel. This update adds a new macro to sychronize the threads' local status with a global status iff there is a local failure on that thread.

### DIFF
--- a/tensorflow_quantum/core/ops/parse_context.h
+++ b/tensorflow_quantum/core/ops/parse_context.h
@@ -16,6 +16,15 @@ limitations under the License.
 #ifndef TFQ_CORE_OPS_PARSE_CONTEXT
 #define TFQ_CORE_OPS_PARSE_CONTEXT
 
+// Syncs a threads work status with some global status.
+#define NESTED_FN_STATUS_SYNC(global_status, local_status, global_lock) \
+  if (TF_PREDICT_FALSE(!local_status.ok())) {                           \
+    global_lock.lock();                                                 \
+    global_status = local_status;                                       \
+    global_lock.unlock();                                               \
+    return;                                                             \
+  }
+
 #include <string>
 #include <vector>
 
--- a/release/BUILD
+++ b/release/BUILD
@@ -40,6 +40,7 @@ sh_binary(
         "//tensorflow_quantum/core/ops:tfq_utility_ops_py",
         "//tensorflow_quantum/core/ops:tfq_simulate_ops_py",
         "//tensorflow_quantum/core/ops/math_ops:inner_product_op_py",
+        "//tensorflow_quantum/core/ops/noise:noisy_samples_op_py",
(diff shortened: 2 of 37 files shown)


tensorflow_quantum/core/ops/parse_context.h
```diff
@@ -16,6 +16,15 @@ limitations under the License.
 #ifndef TFQ_CORE_OPS_PARSE_CONTEXT
 #define TFQ_CORE_OPS_PARSE_CONTEXT
 
+// Syncs a threads work status with some global status.
+#define NESTED_FN_STATUS_SYNC(global_status, local_status, global_lock) \
+  if (TF_PREDICT_FALSE(!local_status.ok())) {                           \
```
jaeyoo, line 21: My first question is about `TF_PREDICT_FALSE`. I saw `global_status` was initialized with `ok`, so I believe that this if block allows `global_status` to be `failure`. So, I expect the `TF_PREDICT_FALSE(!local_status.ok())` means `update if local_status.ok() is not true`. That is, `if (TF_REDICT_TRUE(!local_status.ok())`?

MichaelBroughton: The `TF_PREDICT_TRUE/FALSE` macro is something that hints to the CPUs branch predictor that you expect this condition to be the case almost all of the time. Whenwe have `TF_PREDICT_FALSE(!status.ok())` what that means is tell the branch predictor that you think `!status.ok()` will most of the time be false => status.ok() is true most of the time. In the case of not erroring this is what we would expect. If that is still a little confusing the macro `TF_PREDICT_***` itself doesn't actually change the logic of the code, so if you like you can pretend it's not there :)

tensorflow_quantum/core/ops/parse_context.h
```diff
@@ -16,6 +16,15 @@ limitations under the License.
...
+    global_lock.lock();                                                 \
+    global_status = local_status;                                       \
+    global_lock.unlock();                                               \
+    return;                                                             \
```
zaqqwerty, line 25: Since this is a macro, not a function, should this line be removed?

zaqqwerty (changes requested):
One more question on the macro

tensorflow_quantum/core/ops/parse_context.h
zaqqwerty: Is it intended to break out of the inner loops?

MichaelBroughton: Yes, the macros is intended to break out of the nested function if there was a problem. That's why the return is included.

zaqqwerty (approved):
LGTM

jaeyoo (approved):
LGTM too

PR #521, opened 2021-03-25
fix type mismatch in ip op.

MichaelBroughton:
In the true case this would return type `tf.float32` and in the false case we would get `tf.complex64` updated so it's `tf.complex64` everywhere.


zaqqwerty (approved):
LGTM

PR #522, opened 2021-03-25
Gradient circuits 2/n: LinearCombination

### LINKED ISSUE #514: Implement get_gradient_circuits for LinearCombination

zaqqwerty:
Resolves #514.  The gradient circuits are tested via both direct comparison and via constructing the gradient manually from their expectation values.


tensorflow_quantum/python/differentiators/linear_combination.py
```diff
@@ -92,16 +92,77 @@ def __init__(self, weights, perturbations):
                              "length.")
         if not len(list(set(perturbations))) == len(perturbations):
             raise ValueError("All values in perturbations must be unique.")
-        self.weights = tf.constant(weights)
+        if len(perturbations) < 2:
+            # This is so that tensor squeezing does not cause a problem later.
+            raise ValueError("Must specify at least two perturbations.")
+        self.weights = tf.constant(weights, dtype=tf.float32)
```
MichaelBroughton, line 98: Is this a new error mechanism ? Did we always require the user to have 2 perturbations ?

tensorflow_quantum/python/differentiators/linear_combination.py
```diff
@@ -92,16 +92,77 @@ def __init__(self, weights, perturbations):
...
-        self.perturbations = tf.constant(perturbations)
+        self.perturbations = tf.constant(perturbations, dtype=tf.float32)
+
+        mask = tf.not_equal(self.perturbations,
+                            tf.zeros_like(self.perturbations))
+        self.non_zero_weights = tf.boolean_mask(self.weights, mask)
+        self.zero_weights = tf.boolean_mask(self.weights,
+                                            tf.math.logical_not(mask))
+        self.non_zero_perturbations = tf.boolean_mask(self.perturbations, mask)
+        self.n_non_zero_perturbations = tf.shape(self.non_zero_perturbations)[0]
```
MichaelBroughton, line 108: Can't do `[0]` => `tf.gather(tf.shape(...), 0)`

MichaelBroughton (commented):
These look good. Wondering why we are limiting things down to needing 2 values now ? Couldn't we supply just 1 (in the case of ForwardDifference) ?

tensorflow_quantum/python/differentiators/linear_combination.py
zaqqwerty: Done

tensorflow_quantum/python/differentiators/linear_combination.py
zaqqwerty: Having only one perturbation is the same as evaluating the circuit once at an offset, there need to be at least two evaluation points for differentiation.  There wasn't a limit before; I've now expanded the error message.

In particular, inside ForwardDifference we have `grid_points_to_eval = np.arange(0, error_order + 1)`, and `error_order` is required to be greater than 0, so it also already required 2 or more eval points.  Does this make sense?

MichaelBroughton: Ahhhh that does make sense!

In that case I think all we need is a few tests for these new error mechanisms and we're good to merge.

zaqqwerty: There is a test now checking that at least two perturbations are specified

MichaelBroughton (approved):
LGTM

PR #524, opened 2021-03-29
Added GAD channel support.

MichaelBroughton:


tensorflow_quantum/core/src/circuit_parser_qsim.cc
```diff
@@ -641,6 +641,25 @@ inline Status DepolarizingChannel(const Operation& op,
...
 
+inline Status GADChannel(const Operation& op, const unsigned int num_qubits,
+                         const unsigned int time, NoisyQsimCircuit* ncircuit) {
+  int q;
+  bool unused;
+  float p, gamma;
+  Status u;
+  unused = absl::SimpleAtoi(op.qubits(0).id(), &q);
+
+  u = ParseProtoArg(op, "p", {}, &p);
```
zaqqwerty, line 652: Should have two separate statuses here, currently bad status from parsing p will be ignored

zaqqwerty (changes requested):
Error handling nit

tensorflow_quantum/core/src/circuit_parser_qsim.cc
MichaelBroughton: Good catch. Made consistent with other gates.

zaqqwerty (approved):
LGTM

PR #531, opened 2021-04-03
Add tfq.noise.sampled_expectation.

MichaelBroughton:
Adds support for noisy sample based expectation. Note it's basically identical to the analytical case but we instead call into `ComputeSampledExpectationQsim`. Since we are only ever taking 1 shot per circuit realization I had to modify this function to no longer have a fixed seed.

### DIFF
--- a/tensorflow_quantum/core/ops/noise/noisy_sampled_expectation_op.py
+++ b/tensorflow_quantum/core/ops/noise/noisy_sampled_expectation_op.py
@@ -1,0 +1,64 @@
+# Copyright 2020 The TensorFlow Quantum Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Module for high performance noisy circuit sampled epxectation ops."""
+import os
+import tensorflow as tf
+from tensorflow_quantum.core.ops.load_module import load_module
+
+NOISY_OP_MODULE = load_module(os.path.join("noise", "_tfq_noise_ops.so"))
+
+
+def sampled_expectation(programs, symbol_names, symbol_values, pauli_sums,
+                        num_samples):
+    """Estimate (via sampling) expectation values using monte-carlo simulation.
+
+    Simulate the final state of `programs` given `symbol_values` are placed
+    inside of the symbols with the name in `symbol_names` in each circuit.
+    Channels in this simulation will be "tossed" to a certain realization
+    during simulation. This simulation is repeated `num_samples` times and
+    bitstring based expectation calculations with the given `pauli_sums` are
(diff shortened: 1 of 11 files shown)


tensorflow_quantum/core/ops/noise/noisy_sampled_expectation_op.py
```diff
@@ -0,0 +1,97 @@
...
+import os
+import tensorflow as tf
+from tensorflow_quantum.core.ops.load_module import load_module
+
+NOISY_OP_MODULE = load_module(os.path.join("noise", "_tfq_noise_ops.so"))
+
+
+def sampled_expectation(programs, symbol_names, symbol_values, pauli_sums,
+                        num_samples):
+    """Estimate (via sampling) expectation values using monte-carlo simulation.
```
jaeyoo, line 25: nit: Estimate`s`

tensorflow_quantum/core/ops/noise/tfq_noisy_sampled_expectation.cc
```diff
@@ -0,0 +1,403 @@
...
+      max_num_qubits = std::max(max_num_qubits, num);
+    }
+
+    // Cross reference with standard google cloud compute instances
+    // Memory ~= 2 * num_threads * (2 * 64 * 2 ** num_qubits in circuits)
+    // e2s2 = 2 CPU, 8GB -> Can safely do 25 since Memory = 4GB
+    // e2s4 = 4 CPU, 16GB -> Can safely do 25 since Memory = 8GB
+    // ...
+    if (max_num_qubits >= 26) {
+      // If the number of qubits is lager than 24, we switch to an
```
jaeyoo, line 137: nit: 24(ln 137)? 25(ln133,134)? 26(ln136)?

tensorflow_quantum/core/ops/noise/tfq_noisy_sampled_expectation.cc
```diff
@@ -0,0 +1,403 @@
...
+      const auto tfq_for = qsim::SequentialFor(1);
+      int largest_nq = 1;
+      Simulator sim = Simulator(tfq_for);
+      StateSpace ss = StateSpace(tfq_for);
+      auto sv = ss.Create(largest_nq);
+      auto scratch = ss.Create(largest_nq);
+
+      for (int i = 0; i < ncircuits.size(); i++) {
+        int nq = num_qubits[i];
+        int rep_offset = rep_offsets[start][i];
```
jaeyoo, line 283: no iterations between `start` and `end`. is it intended?

jaeyoo (commented):
I have only nit questions :)

tensorflow_quantum/core/ops/noise/noisy_sampled_expectation_op.py
MichaelBroughton: Fixed.

tensorflow_quantum/core/ops/noise/tfq_noisy_sampled_expectation.cc
MichaelBroughton: Yes, I use a fixed size parallel for block instead of adaptive (like the other ops), so this fine.

tensorflow_quantum/core/ops/noise/tfq_noisy_sampled_expectation.cc
MichaelBroughton: Fixed. 26 is when each thread would need more than 1GB of memory and it makes sense to switch parallel schemes.

jaeyoo (approved):
LGTM

PR #533, opened 2021-04-05
Gradient circuits 4/n: update LinearCombination

zaqqwerty:
For after #532.  During implementation of that PR, I discovered a more efficient way to represent the gradient structure: separating the old `batch_mapper` into a weights tensor and an index tensor.  In the updated tests in this PR, it can be seen that this eliminates the zero padding previously required.

### DIFF
--- a/tensorflow_quantum/python/differentiators/linear_combination.py
+++ b/tensorflow_quantum/python/differentiators/linear_combination.py
@@ -151,11 +151,12 @@ class LinearCombination(differentiator.Differentiator):
         batch_symbol_values = bare_symbol_values + all_perts
 
         # The weights for all the programs.
-        tiled_weights = tf.tile(tf.expand_dims(self.non_zero_weights, 0), [n_symbols, 1])
+        tiled_weights = tf.tile(tf.expand_dims(self.non_zero_weights, 0),
+                                [n_symbols, 1])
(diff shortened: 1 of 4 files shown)


zaqqwerty:
For some reason the tests do not run and instantly fail - the summary says "GitHub Actions has encountered an internal error when running your job."

jaeyoo:
> For some reason the tests do not run and instantly fail - the summary says "GitHub Actions has encountered an internal error when running your job."

it will disappear after several hours. no worries, please take your time :)

tensorflow_quantum/python/differentiators/linear_combination.py
```diff
@@ -111,7 +111,6 @@ def __init__(self, weights, perturbations):
         self.n_non_zero_perturbations = tf.gather(
             tf.shape(self.non_zero_perturbations), 0)
 
-    @tf.function
```
MichaelBroughton, line 114: Wouldn't we want to keep this around ?

MichaelBroughton (commented):
Changes look good. Just one concern on removing the decorator.

tensorflow_quantum/python/differentiators/linear_combination.py
zaqqwerty: Added back, I had removed during debugging and forgot

MichaelBroughton (approved):

PR #534, opened 2021-04-06
Improved randomness sources in parallel ops.

MichaelBroughton:
Over time I was noticing that the randomness sources in our parallel ops weren't behaving very well. This change removes all of the old either constant or not completely thread safe sources of randomness and replaces them with proper random sequence generators from the standard library.


jaeyoo (commented):
I just have a question. Is mt19937 thread-safe?

MichaelBroughton:
A single mt19937  cannot be created once and then used by multiple threads reliably, that is why In the ops I've given each thread it's own `std::mt19937` which is seeded by it's thread ID so there is no risk of issues there.

jaeyoo (approved):

Issue #479, opened 2021-02-11
small improvement to quantum data tutorial


gatorwatt:
#In quantum data tutorial
#Under philosophy that less code is better code, in opening MNIST loading 
#could strike use of np.newaxis
#replace this line
#x_train, x_test = x_train[..., np.newaxis]/255.0, x_test[..., np.newaxis]/255.0
#with
#x_train, x_test = x_train[...]/255.0, x_test[...]/255.0
#and just replace  plt.imshow(x_train[0, :, :, 0])
#with plt.imshow(x_train[0, :, :])
#(the tf.reshape returns same form in both configurations)

#Further down a typo Choose misspelled as Chooe

MichaelBroughton:
Thanks for raising the issue. Would you be interested in opening a PR with a fix ?

gatorwatt:
Sure happy to. Will be my first PR on someone else’s library. To follow.

PR #539, opened 2021-04-12
Issue 479

gatorwatt:
two small improvements:
- removed a np.newaxis application to MNIST data since form returned from tf.reshape is consistent without (updates to cells 4 & 7)
- typo fix replaced "Chooe" with "Choose"


docs/tutorials/quantum_data.ipynb
```diff
@@ -787,14 +782,16 @@
    "source": [
     "Now you have everything you need to re-label the dataset! Now you can consult with the flowchart to better understand how to maximize performance seperation when re-labeling the dataset:\n",
     "\n",
-    "<img src=\"./images/quantum_data_1.png\">\n",
+    "<img src=\"https://github.com/tensorflow/quantum/blob/master/docs/tutorials/images/quantum_data_1.png?raw=1\">\n",
```
MichaelBroughton, line 785: This change was made by our website and should be kept as just `"\n",`

MichaelBroughton (commented):
Thanks for opening the PR, this seems like a lot of changes for a typo fix and the couple of lines of code in #479 . Sometimes colab comes in and auto updates a lot of unnecessary things. Since these changes are relatively small, would you be able to just edit the `.ipynb` file directly so we don't pick up all of these new changes in the diff history? i.e. Only include the changes to line 287 and 613.

gatorwatt:
Thanks just submitted a new PR that only updates lines 184/287 for code and line 613 for typo fix. You were right those other revisions originated from running the notebook in Colab.

Issue #504, opened 2021-03-11
Consistency in Input Definitions


lockwo:
This is what I brought up in the meeting today.

In the API documentation for TFQ the default value of the differentiator input (to PQC, ControlledPQC, Expectation, etc.) is "differentiator=None". I think this clashes with the other interpretations of None in Keras layers and within TFQ. In other cases when the default is None, that is indicative that there nothing for that option, i.e. that option is completely absent in the execution of the operations. E.g. in classical Keras layers default activation is set to None, this (intuitively and programmatically) means that no activation function is applied after results are calculated. When inputs have default values that are required for the operation they are usually specified as such. E.g. in PQC there has to be some initializer for the operations to work, and as such, this default initializer is specified: tf.keras.initializers.RandomUniform(0, 2 * np.pi)

Differentiator is given a None input, but there is actually default differentiators. Previous interpretations might make us believe that no differentiator is used in the operation of the layer when not specified. Internally, when differentiator is not specified (and therefore set to None) this results in [ForwardDifference or Adjoint differentiators](https://github.com/tensorflow/quantum/blob/f9a015a5664dce88ca99f6de85eaf5376ae0cf3a/tensorflow_quantum/python/layers/circuit_executors/expectation.py#L236). The reason for this might be that there are two possible default values depending on the backend. 

Two potential solutions might be as follows. If it is ok to have the same default differentiators for all backends, then simply set the default to be that, like "differentiator=tfq.differentiators.ForwardDifference". If you want to keep the default differentiators different for different backends, then maybe add some discussion below in the documentation about this. Currently all that is said is: "A value for differentiator can also be supplied in the constructor to indicate the differentiation scheme this PQC layer should use." Perhaps changing this to something like (just an idea), "A differentiator is necessary to calculate the gradients for the parameterized circuit. Valid differentiators are of the tfq.differentiators.Differentiator class, a number of which are provided via tfq.differentiators. If a differentiator is not provided the backend default differentiator is used; for qsim this is the Adjoint differentiator and for other backends the ForwardDifference differentiator".

MichaelBroughton:
This makes a lot of sense, I think we should investigate trying to implement this. @zaqqwerty Do you have any thoughts ?

zaqqwerty:
[...]

PR #540, opened 2021-04-12
Added noise support to SampledExpectation.

MichaelBroughton:
Modifies `SampledExpectation` to support noisy and noiseless backends. This also improves on #504 by giving more intuitive meaning to the `backend` parameter. I wasn't able to modify the differentiator default behavior because: https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/  and we must create a new instance for each layer. I remember now why we didn't do this to begin with.

This change still supports legacy calls into the layer with `backend=None` in the same fashion as the old API, so it is not a breaking change. Please review closely.

cc: @lockwo .


jaeyoo (approved):
LGTM

PR #543, opened 2021-04-14
Loosen differentiator test tols.

MichaelBroughton:
The tolerances on these tests was a little too tight on my local machine so reduced them down.


jaeyoo (approved):
LGTM

PR #544, opened 2021-04-15
Added noisy backend to sample layer.

MichaelBroughton:
Like #540 this PR adds the noisy backend to the `Sample` layer and also helps with #504 . These tests also helped me find an issue on our C++ op where if we set n_samples = 1 and n_circuits < n_threads we might encounter race conditions on writes, so there is also a very small tweak to the .cc file as well.


jaeyoo (approved):
LGTM

PR #545, opened 2021-04-20
Added noisy backend to Expectation layer.

MichaelBroughton:
Redefines `backend` behavior for `tfq.layers.Expectation`. `noisy` requires that a value for `repetitions` be provided at a call time. `noiseless` (the default) behaves just like None , while keeping backwards compatability with None so that we can improve on #504 .

### DIFF
--- a/tensorflow_quantum/python/layers/circuit_executors/expectation.py
+++ b/tensorflow_quantum/python/layers/circuit_executors/expectation.py
@@ -222,7 +222,7 @@ class Expectation(tf.keras.layers.Layer):
                 derivative values of given operators_to_measure and circuit,
                 which must inherit `tfq.differentiators.Differentiator` and
                 implements `differentiate_analytic` method. Defaults to None,
-                which uses `linear_combination.ForwardDifference()`. If
+                which uses `tfq.differentiators.ParameterShift()`. If
                 `backend` is also 'noiseless' then default is
                 `tfq.differentiators.Adjoint`.
 
--- a/tensorflow_quantum/python/layers/circuit_executors/expectation.py
+++ b/tensorflow_quantum/python/layers/circuit_executors/expectation.py
@@ -325,8 +325,8 @@ class Expectation(tf.keras.layers.Layer):
                                                    dtype=tf.dtypes.int32)
 
             if reps_need_tile:
-                # Don't tile up if the user gave a python list that was precisely
-                # the correct size to match circuits outer batch dim.
+                # Don't tile up if the user gave a python list that was
+                # precisely the correct size to match circuits outer batch dim.
                 repetitions = tf.tile(repetitions, [circuit_batch_dim, 1])
 
             if not tf.is_tensor(repetitions):
--- a/tensorflow_quantum/python/layers/circuit_executors/expectation.py
+++ b/tensorflow_quantum/python/layers/circuit_executors/expectation.py
@@ -215,7 +215,7 @@ class Expectation(tf.keras.layers.Layer):
             backend: Optional Backend to use to simulate states. Defaults to
                 the 'noiseless' simulator, options include {'noiseless',
                 'noisy'}. In the noisy case a `repetitions` call argument
-                must be provided. Ysers may also specify a preconfigured cirq
+                must be provided. Users may also specify a preconfigured cirq
                 object to use instead, which must inherit
                 `cirq.sim.simulator.SimulatesExpectationValues`.
             differentiator: Optional Differentiator to use to calculate analytic
--- a/tensorflow_quantum/python/layers/circuit_executors/expectation.py
+++ b/tensorflow_quantum/python/layers/circuit_executors/expectation.py
@@ -13,13 +13,16 @@
 # limitations under the License.
 # ==============================================================================
 """A tf.keras.layer that ingests programs and outputs expectation values."""
+import numbers
+
 import numpy as np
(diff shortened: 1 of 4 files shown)


tensorflow_quantum/python/layers/circuit_executors/expectation.py
```diff
@@ -202,23 +205,25 @@ class Expectation(tf.keras.layers.Layer):
...
         simulating a quantum circuit.
 
         Args:
             backend: Optional Backend to use to simulate states. Defaults to
-                the native TensorFlow simulator (None), however users may also
-                specify a preconfigured cirq object to use instead, which must
-                inherit `cirq.sim.simulator.SimulatesExpectationValues`.
+                the 'noiseless' simulator, options include {'noiseless',
+                'noisy'}. In the noisy case a `repetitions` call argument
+                must be provided. Ysers may also specify a preconfigured cirq
```
zaqqwerty, line 218: "Ysers" -> "Users"

tensorflow_quantum/python/layers/circuit_executors/expectation_test.py
```diff
@@ -42,7 +45,8 @@ class ExpectationTest(tf.test.TestCase):
 
     def test_expectation_instantiate(self):
         """Test that Expectation instantiates correctly."""
-        expectation.Expectation()
```
zaqqwerty, line 45: I think we should also keep the default test `expectation.Expectation()`, and also add `expectation.Expectation(backend=None)` to confirm backwards compatibility

tensorflow_quantum/python/layers/circuit_executors/expectation_test.py
```diff
@@ -171,6 +191,69 @@ def test_static_cases(self):
                                   symbol_values=[[0.5]],
                                   operators=test_psum)
 
+    def test_static_cases_nois(self):
```
zaqqwerty, line 194: Need a case somewhere here where `repetitions` is 2D

zaqqwerty, line 194: nit: `nois` -> `noisy`

tensorflow_quantum/python/layers/circuit_executors/expectation_test.py
```diff
@@ -192,28 +275,41 @@ def test_expectation_simple_tf_train(self):
...
         This model will put a qubit in the zero or one state from a random state
         given the input zero or one. This tests the input signature:
         Expectation([input_value_batch]).
         """
+        noisy = backend == 'noisy'
         bit = cirq.GridQubit(0, 0)
         symbols = sympy.symbols('x y z')
-        circuit = _gen_single_bit_rotation_problem(bit, symbols)
+        circuit = _gen_single_bit_rotation_problem(bit, symbols,
+                                                   True if noisy else False)
```
zaqqwerty, line 300: can just be `noisy` instead of `True if noisy else False` ?

tensorflow_quantum/python/layers/circuit_executors/expectation_test.py
```diff
@@ -192,28 +275,41 @@ def test_expectation_simple_tf_train(self):
...
+class ExpectationFunctionalTests(parameterized.TestCase, tf.test.TestCase):
     """Test hybrid/integrated models that include an expectation layer."""
 
-    def test_simple_param_value_input(self):
+    @parameterized.parameters([
+        {
+            'backend': 'noisy'
+        },
+        {
+            'backend': None  # old API usage
```
zaqqwerty, line 286: Should the `cirq.Simulator()` backend also be added to these parameterized tests?  Not sure about adding to one test and not the others

zaqqwerty (changes requested):
Some nits then good to go

tensorflow_quantum/python/layers/circuit_executors/expectation.py
MichaelBroughton: Fixed.

tensorflow_quantum/python/layers/circuit_executors/expectation_test.py
MichaelBroughton: Added.

tensorflow_quantum/python/layers/circuit_executors/expectation_test.py
MichaelBroughton: Good catch, added that case.

MichaelBroughton: Fixed.

tensorflow_quantum/python/layers/circuit_executors/expectation_test.py
MichaelBroughton: It totally can be lol

tensorflow_quantum/python/layers/circuit_executors/expectation_test.py
MichaelBroughton: I looked through the history and this was originally the only test we had that used the simulator and I think I know why: After adding a `cirq.Simulator` case to all the other tests it tripled the runtime of the entire files tests so I think we had originally struck a balance between coverage and runtime, that might be worth keeping.

zaqqwerty (approved):
LGTM

zaqqwerty (approved):
LGTM

PR #546, opened 2021-04-20
Fix CPP style

zaqqwerty:
Some style fixes.  Also added build hints.


MichaelBroughton (approved):
LGTM, thanks for the cleanup. Will merge once CI passes.

PR #547, opened 2021-04-26
Update ***PQC layers to use new backend conventions.

### LINKED ISSUE #504: Consistency in Input Definitions

MichaelBroughton:
Fixes #504 . Sets the default backend to `'noiseless'` for our existing PQC layers and adds errors that will reffer users to the incoming `Noisy***PQC` layers. This comes after the discussions in the weekly sync that it unlike the Expectation and SampledExpectation layers fitting the notion of `noisy` vs `noiseless` into the PQC layers isn't as clean so we will break them up into two seperate layers.

### DIFF
--- a/tensorflow_quantum/python/layers/high_level/controlled_pqc.py
+++ b/tensorflow_quantum/python/layers/high_level/controlled_pqc.py
@@ -86,7 +86,7 @@ class ControlledPQC(tf.keras.layers.Layer):
     to indicate the differentiation scheme this `ControlledPQC` layer
     should use. Here's how you would take the gradients of the
     above example using a `cirq.Simulator` backend (which is slower
-    than `backend=None` which uses C++):
+    than `backend='noiseless'` which uses C++):
 
 
     >>> bit = cirq.GridQubit(0, 0)
@@ -127,7 +127,7 @@ class ControlledPQC(tf.keras.layers.Layer):
                  operators,
                  *,
                  repetitions=None,
-                 backend=None,
+                 backend='noiseless',
                  differentiator=None,
                  **kwargs):
         """Instantiate this layer.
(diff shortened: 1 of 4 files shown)


zaqqwerty (approved):
LGTM

Issue #550, opened 2021-04-28
Sampling with tfq.layers.Sample() produces same outputs


therooler:
**Python** 3.7
**Tensorflow** 2.3.1
**Tensorflow Quantum** 0.4.0

Hi,

I ran into something strange when I was trying to calculate errorbars on the estimators of some observables.
It seems that there are differences between how the `cirq.Simulator()` object handles random numbers compared to how this is done in the default `tfq.layers.Sample()` backend.

**Expected behavior**
Calling `tfq.layers.Sample()` twice on the same circuit within a script should give me different bitstrings.
Launching the same script multiple times where I sample from a circuit should give me different bitstring for each script.

**Actual behaviour**
Multiple scripts executed at different times gave me exactly the same set bitstrings. Also, reinitializing the sampling layer within a script can give the same set of bitstrings depending on what backend one uses.

Here is a minimal working example illustrating the problem:

```python
import tensorflow_quantum as tfq
import tensorflow as tf
import cirq
import numpy as np

CASE = 'tfq'

def sample_generator(circuits_tf, repetitions, nqubits):
    # take tfq backend or cirq backend
    if CASE == 'tfq':
        sample_layer = tfq.layers.Sample()
    elif CASE=='cirq':
        sample_layer = tfq.layers.Sample(cirq.Simulator())
    # create generator from sample_layer outputs, repeat `repetition` times.
    data_x = tf.reshape(sample_layer(circuits_tf, repetitions=repetitions).to_tensor(), (-1, nqubits))
    data_x = tf.unstack(data_x)
    for d in data_x:
        yield d


def main():
    # tf.random.set_seed('supcom')

    nqubits = 4
    nsamples = 100
    batch_size = 10
    # create a simple circuit
    qubits = cirq.GridQubit.rect(1, 4)
    circuit = cirq.Circuit()
    circuit.append(cirq.H.on_each(qubits))

    # convert to tfq tensor
    circuits_tf = tfq.convert_to_tensor([circuit])

    # create data generator
    sample_gen = tf.data.Dataset.from_generator(sample_generator, args=(circuits_tf, nsamples, nqubits),
                                                output_types=tf.int32)
    sample_gen = sample_gen.batch(batch_size)
    sample_gen = sample_gen.prefetch(3 * batch_size)

    # get samples and return them
    samples = []
    for batch in sample_gen:
        samples.append(batch.numpy())
    return samples


if __name__ == '__main__':
    samples_1 = main()
    samples_2 = main()
[...]

PR #551, opened 2021-04-29
Switch to using tf philox random.

MichaelBroughton:
Upgrades or random number generation to TF built in random number generator. xref #550


jaeyoo (approved):
LGTM with nits: why some uses `Rand64`, but others use `Rand32`? Doesn't it require to be equal to `Rand64` to cover up to the future qubit sizes?

MichaelBroughton:
In the cases where `Rand64` is used, qsim requires uint64 and in the case where `Rand32` is used qsim only requires uint32. I have no idea why they need random seeds with 32 bits and 64 bits in other places, but I just made sure to follow the types.

jaeyoo (approved):
LGTM

PR #555, opened 2021-04-29
Fix tutorial

zaqqwerty:
By using a nightly build, the main content of the notebook can stay the same.  Will update once tfq 0.5 is out.


MichaelBroughton (commented):
This is a good fix for now, but I don't think we should be telling people to use the nightly build. Would it be possible to just add two lines to add the missing interface methods instead ?

MichaelBroughton:
I think we should wait for 0.5.0 to push this API change out publicly. Either way, I was thinking you just add:
```python3
def differentiate_analytic
def differentiate_sampled
```
To the custom differentiatior with a comment saying (if you are using higher than 0.5.0 you don't need this) or something.

MichaelBroughton:
This was a temporary fix until 0.5.0 was cut. 0.5.0 has been cut so this is no longer needed. Closing.

PR #557, opened 2021-05-02
QCNN Multi-worker tutorial

verult:
A tutorial for training a QCNN using multiple workers on top of Kubernetes and Google Cloud.

@MichaelBroughton

### DIFF
--- a/docs/tutorials/qcnn_multiworker/Dockerfile
+++ b/docs/tutorials/qcnn_multiworker/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2021 The TensorFlow Quantum Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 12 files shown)


MichaelBroughton:
@lamberta I was thinking for this tutorial it might make sense to put all of these files into the `/research` branch and then create a `.ipynb` notebook with a little more explanation (to go along with the Blog post) that what is inside the REAMDE file that we feature in `/docs/tutorials`. What do you think ?

verult:
Most of the stuff to run in the README are shell commands rather than Python code. Would that be a good fit for an `.ipynb`?

lamberta:
Yeah, this is not the place for Dockerfiles, Makefiles, and Python files—these can't be published as docs on the https://tensorflow.org/quantum website.

This kind of thing is normally what we use the [tensorflow/examples](https://github.com/tensorflow/examples) repo for. And some projects, like TFP, keep a separate [/examples](https://github.com/tensorflow/probability/tree/master/tensorflow_probability/examples) directory (just don't put it in the `/docs` directory). But these are all *maintained* examples—if it goes into the `master` branch, expect PRs for it. If it's for a blog post and not an official example, then creating a separate `research` branch makes sense.

verult:
Got it. The intention is indeed to maintain this example and update this in the future, as improvements in Kubernetes will make the recommended best practice change in the next few months. @MichaelBroughton what do you think about creating an `example/` directory in the TFQ repo, or should this be part of `tensorflow/examples`?

MichaelBroughton:
After doing some thinking on this I think it would be best for now (ant probably the easiest too) to place this PR into the`research` branch as a new directory. That way we can link to it in the blog post and refer any interested users etc. There are links to the research branch in places like the whitepaper and the original launch post so I think it makes a lot of sense.

MichaelBroughton:
Merged #574 so we are safe to close this.

Issue #517, opened 2021-03-24
Remove parallelization logic from python backends


zaqqwerty:
The functions in `batch_util.py` use custom multithreading to speed up calculations done using cirq interfaces.  However, this system causes issues when the cirq objects are actually interfaces to remote simulators or devices.  Since high performance is handled by the default qsim backend, we should remove this multithreading to give more freedom to the cirq backends.

Mints97:
Does the default backend support multithreading? Will it automatically scale with the number of processors on my machine?

zaqqwerty:
The default backend is [qsim](https://github.com/quantumlib/qsim), which automatically supports multithreading.  This issue is just about custom Cirq backends that call out to real hardware, where we let the given backend choose how to parallelize instead of dictating local multithreading (which can overwhelm some quantum computers).

PR #561, opened 2021-05-05
Remove parallelization for Cirq state backend

zaqqwerty:
Removes parallelization of `batch_calculate_state` in favor of a loop over native Cirq interface calls.  xref #517


MichaelBroughton (approved):
LGTM. I think going forward we can also get rid of using thread safe `views` into the underlying arrays and just write slices to them.

PR #562, opened 2021-05-05
Remove parallelization for Cirq sampled expectation backend

zaqqwerty:
Removes parallelization of `batch_calculate_sampled_expectation` in favor of a loop over native Cirq interface calls. xref #517

### DIFF
--- a/tensorflow_quantum/core/ops/batch_util.py
+++ b/tensorflow_quantum/core/ops/batch_util.py
@@ -454,7 +454,7 @@ def batch_calculate_expectation(circuits, param_resolvers, ops, simulator):
 
 def batch_calculate_sampled_expectation(circuits, param_resolvers, ops,
                                         n_samples, simulator):
-    """Compute expectations from sampling circuits using parallel processing.
+    """Compute expectations from sampling a batch of circuits.
 
     Returns a `np.ndarray` containing the expectation values of `ops`
     applied to a specific circuit in `circuits`, given that the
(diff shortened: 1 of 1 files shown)


MichaelBroughton (commented):
I was able to fix the asyncio issue by adding this to the bottom of `TFQPauliSumCollector`
```python3
def collect(self, sampler):
        """Synchronus collect."""
        while True:
            next_job = self.next_job()
            if next_job is None:
                return

            bitstrings = sampler.run(
                next_job.circuit,
                repetitions=next_job.repetitions
            )
            self.on_job_result(next_job, bitstrings)
```
And then modifying the gathering loop to be:
```python3
all_exp_vals = np.ones(shape=(len(circuits), len(ops[0])),
                           dtype=np.float32) * -2

    for c_index, (c, params) in enumerate(zip(circuits, param_resolvers)):
        # (#679) Just ignore empty programs.
        if len(c.all_qubits()) == 0:
            continue
        circuit = cirq.resolve_parameters(c, params)
        for op_index, op in enumerate(ops[c_index]):
            collector = TFQPauliSumCollector(
                circuit, op, samples_per_term=n_samples[c_index][op_index])

            collector.collect(sampler)
            result = collector.estimated_energy().real

            all_exp_vals[c_index][op_index] = result

    return all_exp_vals
```
Does that help on your end ?

zaqqwerty:
Oh ok, if a synchronous collect is ok I will go with this and drop the asyncio usage

MichaelBroughton:
Yeah if we are dropping parallelization in the pure python ops no need to worry about `async` stuff.

PR #566, opened 2021-05-11
Remove parallelization in batch sample.

MichaelBroughton:
Removes parallel calls into python simulators when sampling. Also fixes datatype outputs to match that of that high speed C++ ops (int8).

### DIFF
--- a/tensorflow_quantum/core/ops/batch_util.py
+++ b/tensorflow_quantum/core/ops/batch_util.py
@@ -345,24 +345,24 @@ def batch_calculate_state(circuits, param_resolvers, simulator):
         return empty_ret
 
     biggest_circuit = max(len(circuit.all_qubits()) for circuit in circuits)
+
+    # Default to state vector unless we see densitymatrix.
+    return_mem_shape = (len(circuits), 1 << biggest_circuit)
+    post_process = lambda x: x.final_state_vector
     if isinstance(simulator, cirq.DensityMatrixSimulator):
         return_mem_shape = (len(circuits), 1 << biggest_circuit,
                             1 << biggest_circuit)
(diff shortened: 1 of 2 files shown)


tensorflow_quantum/core/ops/batch_util.py
```diff
@@ -538,30 +412,18 @@ def batch_sample(circuits, param_resolvers, n_samples, simulator):
...
+        if len(c.all_qubits()) == 0:
+            continue
 
-        pool.starmap(_sample_worker_func, input_args)
+        qb_keys = [(q, str(i)) for i, q in enumerate(sorted(c.all_qubits()))]
+        c_m = c + cirq.Circuit(cirq.measure(q, key=i) for q, i in qb_keys)
+        run_c = cirq.resolve_parameters(c_m, resolver)
+        bits = simulator.sample(run_c, repetitions=n_samples)
+        flat_m = bits[[x[1] for x in qb_keys]].to_numpy().astype(np.int8)
+        return_array[batch, :, :len(qb_keys)] = flat_m[:, ::-1]
```
zaqqwerty, line 426: Would it work to do one fewer cycles of reversing: `return_array[batch, :, :len(qb_keys)] = flat_m[:, ::-1]` -> `return_array[batch, :, biggest_circuit - len(qb_keys):] = flat_m`, then just `return return_array`?

zaqqwerty (changes requested):
Thanks for getting this over the finish line for me, one question

tensorflow_quantum/core/ops/batch_util.py
MichaelBroughton: Oops, good call. Done.

zaqqwerty (approved):
LGTM

PR #588, opened 2021-06-07
Duplicate dict keys

FrankFeenix:


FrankFeenix:
@googlebot I signed it!

FrankFeenix:
@MichaelBroughton pls review this

MichaelBroughton:
Hi @FrankFeenix thanks for opening the PR.

Looking at the code here I think what we actually wanted to do was test in the case of `sample_based` being both truth and false and not just False:

```python3
@parameterized.parameters([{'sample_based': True}, {'sample_based': False}])
```

Would you mind updating the code ?

PR #593, opened 2021-06-14
Updated tensorflow.org/quantum landing cards.

MichaelBroughton:
Added new link cards on the landing page. First row: a link to the distributed blog post.
Second row: a link to the research branch of the github.


lamberta (approved):
I'd say two rows of 4 is about the max before we should remove entries. 3 rows is a bit too cluttered, imho

PR #604, opened 2021-07-09
Add notebook from whitepaper to repo

therooler:
Hi, 

As discussed today, I would like to add my notebook for the TFQ whitepaper here so that there is a permanent place for people to check out the code. [Old Colab link](https://colab.research.google.com/drive/1kxUykovB0YMo0gBZd_r-8tJg4pzmO8Lc?usp=sharing).

It's not really readable right now, but I will try to clean this up over the weekend.

Do you need to review these as well?


MichaelBroughton (approved):
LGTM

PR #606, opened 2021-07-10
Quantum embedding

tonybruguier:
This PR tries to re-create a series of paper by Schuld et al. regarding the universality of quantum embeddings

### DIFF
--- a/quantum_embed/quantum_embed.ipynb
+++ b/quantum_embed/quantum_embed.ipynb
@@ -1,0 +1,349 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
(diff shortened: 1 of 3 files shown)


tonybruguier:
I ran some rudimentary profiling. More than happy to receive feedback on this (in addition to code quality, of course):

profile_pie_chart

profile_host_side

profile_tf_ops

MichaelBroughton (changes requested):
This is a good start. Normally in our research branch, we like to see a simple application demonstration in the code. In the case of this embedding, would you mind setting up a simple training loop with a model and some data in order to show off how things work with the Layer ?

tonybruguier:
The unit tests already had a training loop, but it's a good point nonetheless, so I added a Jupyter notebook. PTAL.

MichaelBroughton (approved):

PR #608, opened 2021-07-24
Add Quantum Generative Adversarial Networks tutorial

jaeyoo:


YZNIU (dismissed):

YZNIU (approved):

MichaelBroughton:
I think this notebook may be a fair ways away from being "tutorial ready". We do have a very similar notebook in the `/research` branch which is accessible from the website. I'm not sure if we want to go through the trouble to "tutorialize" this right now. Do you feel strongly we should push to put this in ?

PR #616, opened 2021-08-16
Update load_module.py

### LINKED ISSUE #507: tensorflow quantum giving a issue like this whenever I run it on colab
Screenshot at 2021-03-19 22-21-26
Please help me out with this
import tensorflow as tf

import cirq
import sympy
import numpy as np
import seaborn as sns
import collections

# visualization tools
%matplotlib inline
import matplotlib.pyplot as plt
from cirq.contrib.svg import SVGCircuit 
this happens when i use this

jaeyoo:
Resolves #507

### DIFF
--- a/tensorflow_quantum/core/ops/load_module.py
+++ b/tensorflow_quantum/core/ops/load_module.py
@@ -43,8 +43,7 @@ def load_module(name):
         return load_library.load_op_library(path)
     except:
         path = os.path.join(get_python_lib(prefix="/usr/local"),
-                            "tensorflow_quantum/core/ops",
-                            name)
+                            "tensorflow_quantum/core/ops", name)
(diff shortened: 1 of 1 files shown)


jaeyoo:
```
!pip install tensorflow==2.5.1
!pip install tfq-nightly==0.6.0.dev20210816
```

is tested in the colab.

MichaelBroughton (commented):
I don't think we should try and fix issues with `get_python_lib` ourselves here. It might be better if we found an alternative to `get_python_lib` that worked in more cases. Maybe we just leave it as is ?

mhucka:
I'm closing this due to the age of the report, and, based on the comments above, it seems better not to pursue the changes.

Issue #357, opened 2020-08-20
Support for Projectors in Expectation Calculations


MichaelBroughton:
Right now Cirq does not allow such things as:

```python3
a = cirq.Z(q0) * cirq.Z(q0) + projector(0111011001)
a.expectation_from_wavefunction(a, ...)
```
One must manually write out the projection operators like:
```python3
a = (1 - cirq.Z(q)) / 2 * (1 + cirq.Z(q2)) / 2 . . .
```

Storing such a large paulisum can quickly grow to be prohibitive despite the matrix having just one entry in the end and the post-processing step being very simple to implement on a real device. It would be nice if we could have this feature put into Cirq and then relayed up and supported in TFQ. What do people think ? @zaqqwerty .

MichaelBroughton:
With the additions of https://github.com/quantumlib/Cirq/pull/4364 in the latest release of Cirq. We are now able to work on this issue.

1. Upgrade to the latest version of Cirq.
2. Implement serialization logic for ProjectorSums.
3. Gradually incroporate ProjectorSum functionality in C++ ops incl. gradients.
4. Add Projector sums to fuzz tests for ops.

tonybruguier:
I am interested but I don't want to over-commit and under-deliver. Let me take a look at it and get back to you all.

For context these are the PRs:
https://github.com/quantumlib/Cirq/pull/4331 (ProjectorString)
https://github.com/quantumlib/Cirq/pull/4364 (ProjectorSum)

Only the first PR has made it to the current latest release:
https://github.com/quantumlib/Cirq/releases/tag/v0.12.0

PR #621, opened 2021-08-19
Increment requirement to Cirq to 0.13.1

tonybruguier:
For issue #357 we would want to increase the version of Cirq used to be able to use new objects.


MichaelBroughton:
Oops, looks like: https://github.com/quantumlib/Cirq/pull/4167 might have broken us. We may have to take some extra steps here to be sure and "demote back down" any CCX or CCZ into `ControlledGate(sub_gate=...)` so that things keep working as intended.

MichaelBroughton (commented):
This looks great! One question: Have we looked into this: https://googleapis.dev/python/protobuf/latest/google/protobuf/message.html#google.protobuf.message.Message.SerializeToString ?

Maybe instead of adding these two new functions `program_from_tensor_to_ascii_proto` we could just add an argument to `convert_to_tensor` like `deterministic_proto_serialization` and see if that fixes things ?

tonybruguier:
It worked indeed. Thanks! I named the parameter `deterministic_proto_serialize` because the style guide limits things to 30 characters and chose its default value to be `False` to leave the serialization unchanged by default.

PTAL.

MichaelBroughton (approved):
LGTM

PR #626, opened 2021-09-06
Update concepts.md

MichaelBroughton:


jaeyoo (commented):
Is there a reason why we need to change the word? Isn't it google's quantum supremacy nature paper blog posting, as the URL suggested?

MichaelBroughton:
There was a big debate about “supremacy” being a bad (offensive) word to use and that “beyond-classical” would be a better alternative. It does make things a little confusing when talking about the paper from 2019, but most people agree “beyond-classical” is better. I actually got a few emails from people internally pointing this out here for us to fix too.

jaeyoo (approved):
> Is there a reason why we need to change the word? Isn't it google's quantum supremacy nature paper blog posting, as the URL suggested?

LGTM

PR #649, opened 2022-01-03
Tutorial about quantum advantage paper

tonybruguier:


tonybruguier:
@MichaelBroughton 
As a first round, maybe high-level comments would be helpful (to be sure I am on the right-track). But happy to receive any feedback.

MichaelBroughton (commented):
Good first pass, great to see the logic nailed down. Few high level thoughts:

1. RE: Structure. I think you've got all the basic parts of the tutorial down. But there are a few things missing. RIght now we go Circuit setup, data gathering, model training done. One of the big messages in this paper is that when training the model, the two copy strategy works and even the best possible single copy strategy (based on shadows or anything else) won't work as well. (i.e. if you setup the same pipeline of circuit setup, data gathering and model training, but you don't do bell measurements in a two copy circuit and instead do shadows the model will train poorly) It might be good to bring in some compare + contrast between single copy + two copy by setting up the circuits, drawing data, running training and then seeing the gap in performance.

2. The circuit generation sections still have some of the "ugliness" to them that we needed to make things work on hardware (e.g. using run_sweep to change basis measurements or add randomized X flips). For clarity sake it is probably worthwhile to get rid of this ugliness in favor of keeping things a little simpler and making the code shorter. (this is mostly for sections 1 + 2)
3. It might be good to throw in some motivating text at the beginning and in between sections just to explain what's going on. If you take a look at https://www.tensorflow.org/quantum/tutorials/quantum_data we do have some text walking through the high level ideas + important takeaways.  A diagram might not hurt in this situation either (feel free to recreate + rip off visuals in the paper, but don't directly cut paste them, journals hate that lol).
4. I think it might be fun to try out the new MPS_Sample ops since we are running 1D circuits with such little entanglement we could comfortably simulate ~100 qubits @ bond_dim 8 here without issue and it might be a good chance to show off some of the new library features and talk about them a little bit as well :).

tonybruguier:
Thanks for the high-level review, @MichaelBroughton and sorry for the delay. I didn't understand some of the comments you had so I had to plug some gaps in my knowledge. I also have been somewhat busy, but I'm still eager to continue on this. I tried to address all your high-level comments.

PTAL but it's obviously not urgent.

MichaelBroughton:
Finally got around to re-reading this. Did you get a chance to look into the high level feedback I left before ? I see the tutorial is still missing some comparison visuals (1), simpler code for circuit generation (2) and I don't think we are making the best use of the `mps` ops batching behavior (not making use of outermost batch_dim) (4). What do we think about adding a diagram, maybe copying one from the paper https://arxiv.org/pdf/2112.00778.pdf in lucidchart or something like that (3) ?

PR #568, opened 2021-05-15
Revert "Upgrade to TensorFlow 2.5.0."

MichaelBroughton:
Reverts tensorflow/quantum#567 .
The upgrade to 2.5.0 seems to have caused some pretty serious performance regressions for us. Plan is to back off and cut 0.5.0 depending on 2.4.1 and revisit upgrade at a later date.

PR #650, opened 2022-01-03
Upgrade to TF 2.7.0.

MichaelBroughton:
Supersedes #638 

Also included:
1. Reverted pip install instructions to correct release version of TensorFlow.
2. Added python snippet to top of all test files to remove conflicting proto definitions on python path. I feel like there should be some way to do this in Bazel, but after trying for a few hours I'm out of ideas.


zaqqwerty:
Nice, looking forward to using TF 2.7 features with TFQ!  Past TF version upgrades sometimes caused performance issues (see #568), how are things looking on that front?

MichaelBroughton:
Not sure how performance looks. I think a good way to test will be to compare side-by-side some of the tutorial notebooks and see how much slower/faster the training loops get.

PR #653, opened 2022-01-05
Add SPSA optimizer

lockwo:
SPSA is a common optimizer that plays a prominent role in many quantum optimization procedures. It already exists in other popular packages such as [qiskit](https://qiskit.org/documentation/stubs/qiskit.algorithms.optimizers.SPSA.html#qiskit.algorithms.optimizers.SPSA). Although there are external implementations (e.g. noisyopt), providing a built in version offers a number of advantages (not the least of which is ease of use). The code is short and not complex (and shares the same format as the rotosolve optimizer). I would also like to answer two question which may arise (and I had when implementing):

Why SPSA? 
There are a number of black box optimization algorithms; however, SPSA is probably one of most common in the field of quantum optimization. It is a go to optimizer for VQE (see https://www.nature.com/articles/nature23879) as it can be somewhat robust to noise and take O(1) forward passes independent of the number of parameters (unlike PS or other grad approaches). Additionally, the most common black box optimization package (scipy optimize minimize) does not include it. 

Why as an optimizer? 
Theoretically, the gradient approach SPSA uses could be implemented as a differentiator (just do the gradient estimation as a differentiation approach). However, I think there are two primary drawbacks to this. One, there is well documented knowledge about the hyperparameters to be used in SPSA (see https://ieeexplore.ieee.org/document/705889). If the gradient update was implemented as a differentiator, these would have to be incorporated into a TF optimizer somehow and would likely not be a friendly interface (adjusting how the differentiation is done, e.g. the perturbation parameter, during optimization external to the class doesn't sound good). Additionally, (and this may not actually be true, it is purely based on intuition and some toy examples), the noisy-ness of the gradient estimation (which is fine on average) may be problematic for hybrid systems (e.g. when using the gradients and backproping through) thus the optimizer more readily encourages non-hybrid systems. 

I recognize this is an unsolicited (if you will) PR so if this isn't something that is a desired inclusion it's fine, but I have been using SPSA extensively in qiskit and I think TFQ would benefit from its inclusion. This PR is a simple implementation I drafted this afternoon but there is room for advancements (e.g. second order approximations).

### DIFF
--- a/tensorflow_quantum/python/optimizers/spsa_minimizer.py
+++ b/tensorflow_quantum/python/optimizers/spsa_minimizer.py
@@ -82,10 +82,10 @@ SPSAOptimizerResults = collections.namedtuple(
         'gamma',
         # Specifies scaling of the size of the perturbations
         'blocking',
-        # If true, then the optimizer will only accept updates that improve 
+        # If true, then the optimizer will only accept updates that improve
         # the objective function.
         'allowed_increase'
-        # Specifies maximum allowable increase in objective function 
+        # Specifies maximum allowable increase in objective function
         # (only applies if blocking is true).
     ])
 
@@ -212,14 +212,17 @@ def minimize(expectation_value_function,
                                       maxval=2,
                                       dtype=tf.int32) - 1, tf.float32)
 
-            v_m =  expectation_value_function(state.position - state.c * delta_shift)
-            v_p = expectation_value_function(state.position + state.c * delta_shift)
+            v_m = expectation_value_function(state.position -
+                                             state.c * delta_shift)
+            v_p = expectation_value_function(state.position +
+                                             state.c * delta_shift)
 
             gradient_estimate = (v_p - v_m) / (2 * state.c) * delta_shift
             update = state.a * gradient_estimate
 
             current_obj = expectation_value_function(state.position - update)
-            #state.position.assign(tf.math.floormod(state.position - update, 2* np.pi))
+            # state.position.assign(tf.math.floormod(state.position - 
+            #                                        update, 2* np.pi))
             if state.num_objective_evaluations == 0 or state.objective_value_previous_iteration \
                 < current_obj + state.allowed_increase or not state.blocking:
                 state.position.assign(state.position - update)
--- a/tensorflow_quantum/python/optimizers/spsa_minimizer.py
+++ b/tensorflow_quantum/python/optimizers/spsa_minimizer.py
@@ -223,7 +223,8 @@ def minimize(expectation_value_function,
             current_obj = expectation_value_function(state.position - update)
             # state.position.assign(tf.math.floormod(state.position -
             #                                        update, 2* np.pi))
-            if state.num_objective_evaluations == 0 or state.objective_value_previous_iteration \
+            if state.num_objective_evaluations == 0 or \
+                state.objective_value_previous_iteration \
                 < current_obj + state.allowed_increase or not state.blocking:
                 state.position.assign(state.position - update)
 
--- a/tensorflow_quantum/python/optimizers/spsa_minimizer.py
+++ b/tensorflow_quantum/python/optimizers/spsa_minimizer.py
@@ -194,7 +194,6 @@ def minimize(expectation_value_function,
 
         a_init = tf.convert_to_tensor(a, name='initial_a', dtype='float32')
         c_init = tf.convert_to_tensor(c, name='initial_c', dtype='float32')
-    
 
         def _spsa_once(state):
             """Caclulate single SPSA gradient estimation
@@ -230,7 +229,6 @@ def minimize(expectation_value_function,
 
             return [state]
 
-
         # The `state` here is a `SPSAOptimizerResults` tuple with
         # values for the current state of the algorithm computation.
         def _cond(state):
@@ -241,7 +239,6 @@ def minimize(expectation_value_function,
 
         def _body(state):
             """Main optimization loop."""
-            
             new_a = a_init / (
                 (tf.cast(state.num_iterations + 1, tf.float32) +
                  0.01 * tf.cast(max_iterations, tf.float32))**state.alpha)
--- a/tensorflow_quantum/python/optimizers/spsa_minimizer.py
+++ b/tensorflow_quantum/python/optimizers/spsa_minimizer.py
@@ -260,9 +260,8 @@ def minimize(expectation_value_function,
             return [state]
 
         initial_state = _get_initial_state(initial_position, tolerance,
-                                           expectation_value_function, a,
-                                           alpha, c, gamma, blocking,
-                                           allowed_increase)
+                                           expectation_value_function, a, alpha,
+                                           c, gamma, blocking, allowed_increase)
 
         initial_state.objective_value.assign(
             expectation_value_function(initial_state.position))
--- a/tensorflow_quantum/python/optimizers/spsa_minimizer.py
+++ b/tensorflow_quantum/python/optimizers/spsa_minimizer.py
@@ -78,12 +78,12 @@ SPSAOptimizerResults = collections.namedtuple(
         # Specifies the learning rate
         'alpha',
         # Specifies scaling of the learning rate
-        'c', 
+        'c',
         # Specifies the size of the perturbations
         'gamma',
         # Specifies scaling of the size of the perturbations
         'blocking',
-        # If true, then the optimizer will only accept updates that improve the objective function. 
+        # If true, then the optimizer will only accept updates that improve the objective function.
         'allowed_increase'
         # Specifies maximum allowable increase in objective function (only applies if blocking is true).
     ])
@@ -206,7 +206,11 @@ def minimize(expectation_value_function,
             Returns:
                 states: A list which the first element is the new state
             """
-            delta_shift = tf.cast(2 * tf.random.uniform(shape=state.position.shape, minval=0, maxval=2, dtype=tf.int32) - 1, tf.float32)
+            delta_shift = tf.cast(
+                2 * tf.random.uniform(shape=state.position.shape,
+                                      minval=0,
+                                      maxval=2,
+                                      dtype=tf.int32) - 1, tf.float32)
 
             v_m, v_p =  expectation_value_function(state.position - state.c * delta_shift),\
                  expectation_value_function(state.position + state.c * delta_shift)
@@ -238,8 +242,11 @@ def minimize(expectation_value_function,
         def _body(state):
             """Main optimization loop."""
             
-            new_a = a_init / ((tf.cast(state.num_iterations + 1, tf.float32) + 0.01 * tf.cast(max_iterations, tf.float32))**state.alpha)
-            new_c = c_init / (tf.cast(state.num_iterations + 1, tf.float32)**state.gamma)
+            new_a = a_init / (
+                (tf.cast(state.num_iterations + 1, tf.float32) +
+                 0.01 * tf.cast(max_iterations, tf.float32))**state.alpha)
+            new_c = c_init / (tf.cast(state.num_iterations + 1, tf.float32)**
+                              state.gamma)
 
             state.a.assign(new_a)
             state.c.assign(new_c)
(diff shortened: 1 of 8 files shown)


lockwo:
I see there are some formatting problems, I will fix those shortly.

lockwo:
Sorry there are so many commits, a bit of my github/TF testing naivete showing. 99% of the them are tiny formatting adjustments.

tensorflow_quantum/python/optimizers/spsa_minimizer.py
```diff
@@ -0,0 +1,275 @@
...
+    tf.Tensor(-4.7045116, shape=(), dtype=float32)
+
+    Args:
+        expectation_value_function:  A Python callable that accepts
+            a point as a real `tf.Tensor` and returns a `tf.Tensor`s
+            of real dtype containing the value of the function.
+            The function to be minimized. The input is of shape `[n]`,
+            where `n` is the size of the trainable parameters.
+            The return value is a real `tf.Tensor` Scalar (matching shape
+            `[1]`).  This must be a linear combination of quantum
```
MichaelBroughton, line 154: Can we tighten up the wording here to something like: 

"Python callable that accepts a real valued `tf.Tensor` with shape [n] where n is the number of function parameters,....."

tensorflow_quantum/python/optimizers/spsa_minimizer.py
```diff
@@ -0,0 +1,275 @@
...
+        "gamma": tf.Variable(gamma),
+        "blocking": tf.Variable(blocking),
+        "allowed_increase": tf.Variable(allowed_increase)
+    }
+    return SPSAOptimizerResults(**init_args)
+
+
+def minimize(expectation_value_function,
+             initial_position,
+             tolerance=1e-7,
```
MichaelBroughton, line 116: Might want to lower this to 1e-4 or 1e-5. Float32's aren't reliable to this many decimals.

tensorflow_quantum/python/optimizers/spsa_minimizer.py
```diff
@@ -0,0 +1,275 @@
...
+             max_iterations=60,
+             alpha=0.602,
+             a=1.0,
+             c=1.0,
```
MichaelBroughton, line 120: Instead of calling these "a" and "c", would we be able to do something like "lr" or "lr_scaling" just to be a little more descriptive names ?

tensorflow_quantum/python/optimizers/spsa_minimizer.py
```diff
@@ -0,0 +1,275 @@
...
+    >>> n = 10  # Number of sinusoids
+    >>> coefficient = tf.random.uniform(shape=[n])
+    >>> min_value = -tf.math.reduce_sum(tf.abs(coefficient))
+    >>> func = lambda x:tf.math.reduce_sum(tf.sin(x) * coefficient)
+    >>> # Optimize the function with SPSA, start with random parameters
+    >>> result = tfq.optimizers.SPSA_minimize(func, np.random.random(n))
+    >>> result.converged
+    tf.Tensor(True, shape=(), dtype=bool)
+    >>> result.objective_value
+    tf.Tensor(-4.7045116, shape=(), dtype=float32)
```
MichaelBroughton, line 145: Can we add a unit test that covers the use case shown here ?

tensorflow_quantum/python/optimizers/spsa_minimizer.py
```diff
@@ -0,0 +1,275 @@
...
+            v_p = expectation_value_function(state.position +
+                                             state.c * delta_shift)
+
+            gradient_estimate = (v_p - v_m) / (2 * state.c) * delta_shift
+            update = state.a * gradient_estimate
+
+            current_obj = expectation_value_function(state.position - update)
+            # state.position.assign(tf.math.floormod(state.position -
+            #                                        update, 2* np.pi))
+            if state.num_objective_evaluations == 0 or \
```
MichaelBroughton, line 226: Nit: Remove comment

tensorflow_quantum/python/optimizers/spsa_minimizer.py
```diff
@@ -0,0 +1,275 @@
...
+
+    Here is an example of optimize a function which consists summation of
+    a few sinusoids.
+
+    >>> n = 10  # Number of sinusoids
+    >>> coefficient = tf.random.uniform(shape=[n])
+    >>> min_value = -tf.math.reduce_sum(tf.abs(coefficient))
+    >>> func = lambda x:tf.math.reduce_sum(tf.sin(x) * coefficient)
+    >>> # Optimize the function with SPSA, start with random parameters
+    >>> result = tfq.optimizers.SPSA_minimize(func, np.random.random(n))
```
MichaelBroughton, line 141: Nit: spsa_minimize

tensorflow_quantum/python/optimizers/spsa_minimizer_test.py
```diff
@@ -0,0 +1,153 @@
...
+        loss_value = loss(model(train_x, training=True), train_y)
+        return loss_value
+
+    return func
+
+
+class SPSAMinimizerTest(tf.test.TestCase, parameterized.TestCase):
+    """Tests for the SPSA optimization algorithm."""
+
+    def test_nonlinear_function_optimization(self):
```
MichaelBroughton, line 90: Could we add three (or more) tests with simple learning functions (maybe more sinusoids?) to verify a few other key baseline functionalities:

1. If failed to converge, does the number of function iterations in the result line up with the supplied arguments
2. If failed to converge due to tolerance misses, is that in fact reflected in a case where convergence to the tolerance is very close
3. do blocking and allow_increase interact correctly with one another and produce expected results when they are enabled/disabled and changed.

tensorflow_quantum/python/optimizers/spsa_minimizer_test.py
```diff
@@ -0,0 +1,153 @@
...
+        """Test to optimize a non-linear function.
+        """
+        func = lambda x: x[0]**2 + x[1]**2
+
+        result = spsa_minimizer.minimize(func, tf.random.uniform(shape=[2]))
+        print(func(result.position))
+        self.assertAlmostEqual(func(result.position).numpy(), 0, delta=1e-4)
+        self.assertTrue(result.converged)
+
+    def test_keras_model_optimization(self):
```
MichaelBroughton, line 100: Nice job fuzzing our a more complex use case.

MichaelBroughton (changes requested):
Great first pass! I think just a few tweaks here and there along with some more baseline test coverage to ensure the namedtuple results are behaving correctly and we should be good to merge.

tensorflow_quantum/python/optimizers/spsa_minimizer.py
lockwo: Changed

tensorflow_quantum/python/optimizers/spsa_minimizer.py
lockwo: Changed to 1e-5

tensorflow_quantum/python/optimizers/spsa_minimizer.py
lockwo: Changed

tensorflow_quantum/python/optimizers/spsa_minimizer.py
lockwo: Changed to lr and perturb, respectively. Added seed parameter.

tensorflow_quantum/python/optimizers/spsa_minimizer.py
lockwo: I modified the example and added the same format test case

tensorflow_quantum/python/optimizers/spsa_minimizer_test.py
lockwo: I added tests for 1, 3 but I'm not sure what you mean by 2. Like it fails to converge because it evaluates two times and the different is smaller than the tolerance?

tensorflow_quantum/python/optimizers/spsa_minimizer_test.py
lockwo: It's a lot smaller tolerance than the standard implementation :) (see https://github.com/andim/noisyopt/blob/master/noisyopt/tests/test_noisyopt.py#L97). I compared with existing implementations. Additionally, I added a test case that is basically just the one from qiskit (https://github.com/Qiskit/qiskit-terra/blob/main/test/python/algorithms/optimizers/test_spsa.py) and they have <= -0.95 when the minimum is -1 (which it doesn't reach even with exact expectation values), so it doesn't seem uncommon.

lockwo:
How is "//tensorflow_quantum/python/differentiators:gradient_test" failing when I changed 3 unrelated files? It wasn't failing before.

MichaelBroughton:
unfortunately It's a little flaky (it's on the todo list to fix). Have you got things in a good start where you'd like another review ?

lockwo:
I addressed most things, if you could just go over my conversation responses I had a question or two and with those answered I'll fix everything up and then be ready for round 2.

tensorflow_quantum/python/optimizers/spsa_minimizer.py
```diff
@@ -0,0 +1,276 @@
...
+            is used.
+
+    Returns:
+        optimizer_results: A SPSAOptimizerResults object contains the
+            result of the optimization process.
+    """
+
+    with tf.name_scope(name or 'minimize'):
+        if seed is not None:
+            tf.random.set_seed(seed)
```
MichaelBroughton, line 185: This is setting the global seed, which we don't want (it will impact code outside of this function). Something like this might be a little better:

```python3
g = tf.random.Generator.from_seed(1234)
g.normal(shape=(2, 3)) # produces same output as tf.random.normal, but from the seed in g.
```

tensorflow_quantum/python/optimizers/spsa_minimizer.py
```diff
@@ -0,0 +1,276 @@
...
+            point, or points when using batching dimensions, of the search
+            procedure. At these points the function value and the gradient
+            norm should be finite.
+        tolerance: Scalar `tf.Tensor` of real dtype. Specifies the tolerance
+            for the procedure. If the supremum norm between two iteration
+            vector is below this number, the algorithm is stopped.
+        a: Scalar `tf.Tensor` of real dtype. Specifies the learning rate
+        alpha: Scalar `tf.Tensor` of real dtype. Specifies scaling of the
+            learning rate.
+        c: Scalar `tf.Tensor` of real dtype. Specifies the size of the
```
MichaelBroughton, line 165: Nit: Need to update the names here.

tensorflow_quantum/python/optimizers/spsa_minimizer.py
```diff
@@ -0,0 +1,276 @@
...
+            perturbations.
+        gamma: Scalar `tf.Tensor` of real dtype. Specifies scaling of the
+            size of the perturbations.
+        blocking: Boolean. If true, then the optimizer will only accept
+            updates that improve the objective function.
+        allowed_increase: Scalar `tf.Tensor` of real dtype. Specifies maximum
+            allowable increase in objective function (only applies if blocking
+            is true).
+        name: (Optional) Python `str`. The name prefixed to the ops created
```
MichaelBroughton, line 174: Nit: Need to add seed.

MichaelBroughton (commented):
This is looking pretty good now, just a few more nits and we should be good to merge.

tensorflow_quantum/python/optimizers/spsa_minimizer.py
lockwo: Added

tensorflow_quantum/python/optimizers/spsa_minimizer.py
lockwo: Updated

tensorflow_quantum/python/optimizers/spsa_minimizer.py
lockwo: Updated. There is only 1 random op (for the deltas), but it now draws from the generator.

lockwo:
Alright, I think I made all the updates (and the gradient tests didn't flake this time)

MichaelBroughton (approved):
LGTM

PR #678, opened 2022-04-06
Use run_batch in cirq.Sampler

Cynocracy:
This allows for more efficient implementations of running a batch (for example parallelization).

run_batch delegates to run_sweep in the default implementation, so this should be drop-in equivalent, AIUI.


Cynocracy:
@zaqqwerty :wave: Saw your TODO and think it should be possible now, what do you think?

zaqqwerty:
Yeah this seems to clear up that TODO, thanks for taking care of it!

zaqqwerty (approved):
LGTM

PR #683, opened 2022-04-11
Fixing small documentation bug in tfq.layers.Sample

lockwo:
The CNOT gate in the first example in tfq.layers.Sample has only 1 input (but needs 2)
image

This just adds the other input so it matches the diagram


zaqqwerty (approved):
LGTM

PR #727, opened 2022-11-12
Add new project in 6G for Quantum on AI

emirhanai:
Add new project in 6G for Quantum on AI by Emirhan BULUT

### DIFF
--- a/docs/tutorials/Quantum_with_6G_Technology_in_Computer_Vision_on_AI.ipynb
+++ b/docs/tutorials/Quantum_with_6G_Technology_in_Computer_Vision_on_AI.ipynb
@@ -1,0 +1,1253 @@
+{
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "xLOXFOT5Q40E"
+      },
+      "source": [
+        "##### Copyright 2022 The Emirhan BULUT.\n",
+        "# Quantum with 6G Technology in Computer Vision on AI\n",
+        "**Author:** [Emirhan BULUT](https://www.linkedin.com/in/artificialintelligencebulut/)<br>\n",
+        "**Date created:** 2022/10/31<br>\n",
+        "**Last modified:** 2022/10/31<br>\n",
+        "**Description:** Processed with 2nd class land use image datasets accompanied by quantum neural network in a manner compatible with 6G with quantum computer and compared with CNN (at close parameters)."
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "i9Jcnb8bQQyd"
+      },
+      "source": [
+        "<table class=\"tfo-notebook-buttons\" align=\"left\">\n",
+        "\n",
+        "  <td>\n",
+        "    <a target=\"_blank\" href=\"https://colab.research.google.com/drive/1yS5W-EsBDc6RYGYvypveGCv0QaTBCZxo?usp=sharing\"><img src=\"https://www.tensorflow.org/images/colab_logo_32px.png\" />Run in Google Colab</a>\n",
+        "  </td>\n",
+        "  <td>\n",
+        "    <a target=\"_blank\" href=\"https://github.com/emirhanai\"><img src=\"https://www.tensorflow.org/images/GitHub-Mark-32px.png\" />View source on GitHub</a>\n",
+        "  </td>\n",
+        "</table>"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "udLObUVeGfTs"
+      },
+      "source": [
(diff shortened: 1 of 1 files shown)


lockwo:
How is this meaningfully different from https://www.tensorflow.org/quantum/tutorials/mnist? Also there is no explanation as to what 6G even is.

emirhanai:
> How is this meaningfully different from https://www.tensorflow.org/quantum/tutorials/mnist? Also there is no explanation as to what 6G even is.

Hey hello,
thanks your reply!
So, you know, 6G is the latest technology used right now. I also set an algorithm status close to 6G (I designed the data accordingly). This is the situation that leaves MNIST completely. Pictures are optimized with tensorflow lanczos5 algorithms according to 6G technology. In addition, I processed it in the form of a 4x4 image in line with my possibilities. When processed as 128x128 in a real quantum computer, the F1 coefficient will naturally increase as the speed efficiency will increase thanks to 6G. (The unit time processing coefficient of the speed in memory).

Thanks for this nice issue!

lockwo:
My first question meant, what TFQ features/functionality is different that the MNIST example? It is a downsampled image that is encoded via X gates that is then compared to a CNN. The overall flow and TFQ tooling used seems very similar. It also remains unclear what 6G actually is. As far as I know 5G is a set of telco standards. Is 6G also a set of standards? Is it already used and adopted by the telco industry? How is "lanczos5 algorithms according to 6G technology", lanczos is an algorithm from the 70s right?

emirhanai:
Yes, but lanczos 5 is a new algorithm. Lanczos 5v6 is used in image processing software compatible with the 5G version in 2022.
For example, the derivative was found 200 years ago, but now it is used by describing quantum algorithms in the 2022 version.

If I come to the question, yes, the skeleton is similar, this version is 1. I am currently working on version 2. Here, I will prepare the model parameterically with a syntax suitable for adding batch normalization (in version 2).

The main difference here is the preparation of the data. It is prepared according to MNIST 3G technology. The data here is prepared according to the 6G system.

Finally, 6G is not a set of standards. It is a reflection technology that creates a batch for quantum attacks and quantum machine learning algorithms (quantum physics 2 compatible). For this reason, its speed is 1 Tbps on average.

The basic prediction here is this: The direction of the matrices and the vector basis that 6G processes are directly proportional to the way the data is prepared. For this reason, it is associated with the qubit algorithm sequence to be used.
If 4x4 was used, the algorithm is likely to perform poorly (if you measure on a computer connected to 6G).

The subject is long. Cheers!

lockwo:
It is important for tutorials to be highly explanatory (both of TFQ and of QML), so all the information that you are telling me should probably be in the tutorial (since I am somewhat in the know of QML and so people who are just learning are going to have even more questions than me).

> Yes, but lanczos 5 is a new algorithm. Lanczos 5v6 is used in image processing software compatible with the 5G version in 2022. For example, the derivative was found 200 years ago, but now it is used by describing quantum algorithms in the 2022 version.

What is different from lanczos 1 to 5? I think it is important to define these words to, because when you say "5G" I (and many) will think of the cell network/telco standards 5G (https://en.wikipedia.org/wiki/5G) which is different from image algorithms. So differentiating the common understanding is important. 

> If I come to the question, yes, the skeleton is similar, this version is 1. I am currently working on version 2. Here, I will prepare the model parameterically with a syntax suitable for adding batch normalization (in version 2).

Will the batch norm be in the quantum model? 

> The main difference here is the preparation of the data. It is prepared according to MNIST 3G technology. The data here is prepared according to the 6G system.

How is MNIST 3G? It's just a collection of handwritten data that the government had from the 90s. Also the data preparation is all not TFQ (but normal TF) so that isn't as important for tutorials on TFQ. 

> Finally, 6G is not a set of standards. It is a reflection technology that creates a batch for quantum attacks and quantum machine learning algorithms (quantum physics 2 compatible). For this reason, its speed is 1 Tbps on average.

Citations will probably be important here and could probably resolve a lot of this, if you have papers to read people can learn a lot. The terminology is unclear, what is a "reflection technology"? Even if I try to look online, I cannot find anything about (just some companies named "reflection technology). What is quantum physics 2? 

> The basic prediction here is this: The direction of the matrices and the vector basis that 6G processes are directly proportional to the way the data is prepared. For this reason, it is associated with the qubit algorithm sequence to be used.
If 4x4 was used, the algorithm is likely to perform poorly (if you measure on a computer connected to 6G).

What is the direction of a matrix? That isn't a linear algebra term I am familiar with (outside of rotation matrices). And 6G processes it? 6G is a specific thing, not a class of technology? 

> Finally if you want you can check my short presentation for 6G in Quantum:
https://github.com/emirhanai/Quantum-with-6G-Technology-in-Computer-Vision-on-AI/blob/main/quantum6%20and%20me.pdf

Going through this presentation brings up more questions, it references 5G as IMT-2020 which is a set of standards (https://en.wikipedia.org/wiki/IMT-2020) but now 6G isn't?

PR #740, opened 2023-01-19
Force old version of numpy in `noise.ipynb`

markmcd:
This currently works in colab, but will break when it's updated, and breaks in the TF docs publishing infra.

The TFQ code is referencing NumPy's `np.float`, which is removed in v1.24. [Details here](https://numpy.org/doc/stable/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated). This is quick hack to get the docs working again.

### DIFF
--- a/docs/tutorials/noise.ipynb
+++ b/docs/tutorials/noise.ipynb
@@ -106,7 +106,7 @@
         "id": "J2CRbYRqrLdt"
       },
       "source": [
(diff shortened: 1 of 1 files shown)


jaeyoo:
https://github.com/tensorflow/quantum/pull/749 solves this PR together. closed.

PR #827, opened 2024-12-09
Fix warnings "target is both a rule and a file"

mhucka:
Bazel 6.5.0 complains about

```
(04:17:26) WARNING: ..../external/local_config_tf/BUILD:8216:8:
target 'libtensorflow_framework.so' is both a rule and a file;
please choose another name for the rule
(04:17:26) WARNING: ..../external/local_config_tf/BUILD:8226:8:
target 'test_log_pb2.py' is both a rule and a file;
please choose another name for the rule
```

(Note: I elided the long pathnames in the messages above by replacing them with '....'.) The warnings come from definitions in `third_party/tf/tf_configure.bzl` where the names of a couple of rules are `libtensorflow_framework.so` and `test_log_pb2.py`. Changing the names to `libtensorflow_framework_so` and `test_log_pb2_py` resolves the problem. Other names are possible, of course, but those seem like good choices in order to keep their meanings clear.

### DIFF
--- a/third_party/tf/tf_configure.bzl
+++ b/third_party/tf/tf_configure.bzl
@@ -195,7 +195,7 @@ def _tf_pip_impl(repository_ctx):
         repository_ctx,
         None,
         "",
-        "libtensorflow_framework.so",
+        "libtensorflow_framework_so",
         [tf_shared_library_path],
         ["_pywrap_tensorflow_internal.lib"  if _is_windows(repository_ctx) else "libtensorflow_framework.so"],
     )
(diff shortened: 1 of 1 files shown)


MichaelBroughton (approved):
I think having the name `libtensorflow_framework.so` is helpful for static objects that get rolled into the wheel file. I'm not 100% confident that this name change can be done, but lets try it and see if wheel releases break or not :).

mhucka:
I didn't make this very clear, but the _file_ names are unchanged (i.e., retain `.so` extensions). This _only_ changes the name of a couple of rules defined in the Bazel configuration.

FWIW, I did test the result using TFQ's test (via "bazel test //tensorflow_quantum/..."), after building and installing the TFQ wheel built with these changes into the python virtual environment, and the tests passed. I can do other tests if you have suggestions?

PR #831, opened 2024-12-12
Fix many compiler warnings

mhucka:
This fixes a large number of Bazel build and compiler warnings. Summary:

- Fix various compiler warnings about type mismatches. Pretty much all of these concerned loop variables; the warnings concerned `int`'s being used where unsigned int or vector size types should be used instead, to match the type of the value the loop variable was being compared against.
- Fix compiler warnings about unused variables in `tensorflow_quantum/core/src/circuit_parser_qsim.cc`. These involved status return values from Abseil function calls. I replaced the code with simple testing of the return
values, similar to how it was already done in one place in the file, and in a way that follows the recommendations from the Abseil docs.
- Pull out the definition of function `BalanceTrajectory` from `util_qsim.h` into a separate file (`util_balance_trajectory.cc`), and make adjustments in the BUILD file and files that used the function. This was to eliminate repeated warnings about unused function `BalanceTrajectory` in files that included `util_qsim.h`. Also moved the relevant tests from `util_qsim_test.cc` to their own file.
- Fix a "unused lambda capture" compiler warning in one file.
- Fix a couple of sha256 hashes in the Bazel WORKSPACE file.
- Add Bazel build options to `configure.sh` to silence warnings we can't do anything about because they're coming from packages included as a result of importing TensorFlow Bazel workspace files. The options are scoped as narrowly as possible to limit the impact of hiding warnings to specific cases.

I tested these changes with both gcc 14.2.0 and clang 17.0.6 on a Debian 6.10 Linux system.


MichaelBroughton (changes requested):
This is pretty big and contains a lot of changes that don't have a lot to do with each other. Could we split this into a few smaller PRs ? Perhaps something like:

1. The notebook format as it's own PR.
2. The C++ type changes and some warning fixes.
3. The refactoring of trajectory balancing.
4. The introduction of the additional error checking in some of the parse code.

As it stands there is a lot going on in each of these areas and I worry if we miss things and require a rollback, we also won't have a commit history that lets us separate out things easily.

mhucka:
@MichaelBroughton You're right; this needs to be split up. Making it one big PR was misguided.

New PRs coming up shortly.

PR #835, opened 2025-01-04
Add bazel flags to silence compiler warnings we can't do anything about

mhucka:
A number of harmless warnings during compilation come from third-party things pulled in by the TFQ build rules. They involve things like unused functions, unused variables, or some deprecated constructs. The best way to fix them would be in the original source code, but in some cases this is impossible; for example, we can't change what TensorFlow 2.15.0 needs. The second best way would be to add patch rules in TFQ to adjust the code or the build rules. I spent a couple of hours trying to figure out how to that for the copy of zlib included by bazel_rules, but due to my still-limited Bazel knowledge, I wasn't able to get the patching scheme working. So, for the time being, and in consideration of the harmless nature of the warnings, the approach taken here is to add compile-time per-file flags that tell the compiler not to warn about the specific kinds of problems that we know exist in the files involved.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -102,6 +102,28 @@ write_to_bazelrc "build -c opt"
 write_to_bazelrc "build --cxxopt=\"-D_GLIBCXX_USE_CXX11_ABI=1\""
 write_to_bazelrc "build --cxxopt=\"-std=c++17\""
 
+# The transitive inclusion of build rules from TensorFlow ends up including
+# and building two copies of zlib (one from bazel_rules, one from the TF code
+# baase itself). The version of zlib you get (at least in TF 2.15.0) ends up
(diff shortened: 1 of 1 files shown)


MichaelBroughton (approved):
I'm fine with this, but I think we will be taking on a losing battle as we jump through versions of dependencies, new warnings will undoubtedly arise and old ones will undoubtedly disappear. If it really bothers you I'm fine with this, but we don't want to be making too much work for ourselves here.

mhucka:
Yeah, you're right; this is a path of diminishing returns. For now, I mainly wanted to silence as many warnings as possible in order to make it easier to see new warnings coming from updating dependencies, updated platforms, gpu vs cpu, etc. I will try not to go gonzo with this, though :-).

PR #837, opened 2025-01-04
Refactor to move BalanceTrajectory to separate file

mhucka:
Previously, util_qsim.h included the definitions of the function `BalanceTrajectory`. This led to compiler warnings about undefined functions in places that included util_qsim.h but didn't use all variants of `BalanceTrajectory`.

This refactor simply moves `BalanceTrajectory` to its own file (util_balance_trajectory.cc), and also moves the relevant tests from util_qsim_test.cc to the corresponding util_balance_trajectory.cc.


MichaelBroughton (approved):
LGTM

PR #838, opened 2025-01-04
Fix warnings about unused variables in circuit_parser_qsim.cc

mhucka:
This file produces many warnings of the form

```
tensorflow_quantum/core/src/circuit_parser_qsim.cc:185:8:
warning: variable 'unused' set but not used [-Wunused-but-set-variable]
  185 |   bool unused = absl::SimpleAtoi(op.qubits(0).id(), &q0);
      |        ^
```

Inspecting the code reveals that variable `unused` is indeed set but never used. This construct was used in many places in this file, and always involving `SimpleAtoi`.

I replaced the code with a simple alternative that tests the returned value from `SimpleAtoi`, in a way that mirrors what was already done in one place in the file where `unused` was not used. This approach follows the recommendations from the Abseil docs, too.

Since this change means that new exceptions may be returned if the status from `SimpleAtoi` is not okay, it also needed the addition of test cases in `circuit_parser_qsim_test.cc`.

A final change is to fix a couple of warnings involving int versus unsignned int.


MichaelBroughton (changes requested):
Thanks for the PR. Looking at https://github.com/tensorflow/quantum/blob/4e0c1b7e03d6f1ec61f1c6024853f1d637993c1e/tensorflow_quantum/core/src/program_resolution.h#L38

I remember our process for getting to qsim circuits is this:

1. Replace Grid/Line Qubit representations in the circuit and paulisum protos with integer indices that all line up and are well ordered across the batch.
2. Forward this partially processed proto here to unpack into qsim circuit structs.
3. Actually simulate the circuit placing results back into tensors in the right order.

So whenever I call into any of these gate constructors it is a requirement that all the proto messages contain integers. I might be more inclined to use assertions (or nothing at all and just (void) it out) here.

mhucka:
> Thanks for the PR. Looking at
> https://github.com/tensorflow/quantum/blob/4e0c1b7e03d6f1ec61f1c6024853f1d637993c1e/tensorflow_quantum/core/src/program_resolution.h#L38
> 
> I remember our process for getting to qsim circuits is this:
> 
> 1. Replace Grid/Line Qubit representations in the circuit and paulisum protos with integer indices that all line up and are well ordered across the batch.
> 2. Forward this partially processed proto here to unpack into qsim circuit structs.
> 3. Actually simulate the circuit placing results back into tensors in the right order.
> 
> So whenever I call into any of these gate constructors it is a requirement that all the proto messages contain integers. I might be more inclined to use assertions (or nothing at all and just (void) it out) here.

Oh, I understand now: due to the other steps around these calls to SimpleAtoi, the calls will always get valid values, therefore the returns don't need to be tested.

I lean towards using void for these cases, mainly due to having had the principle of "assert is for test code, not production code" drilled into me for too long …

MichaelBroughton (approved):
LGTM

PR #839, opened 2025-01-04
Fix type mismatch warnings

mhucka:
Many files produced many type mismatch warnings with both clang 17 and gcc 14 on a Debian 6.10.11 x86_64 system. This commit fixes almost all the ones I'm seeing in the TFQ code base.

Some other warnings still arise in other code and modules. Those warnings don't appear to be things we can easily work around short of creating a lot of patch files (at significant cost in time and testing). For those, I'm currently relying on setting compiler options (e.g., `-Wno-unused-function` for warnings about unused functions) narrowed in scope to the problematic package using Bazel's `--per_file_copt` and `--host_per_file_copt` flags. Those flags are not in this PR.

Note: this PR should be applied _after_ PR #837.


tensorflow_quantum/core/ops/math_ops/tfq_inner_product_grad.cc
```diff
@@ -62,8 +62,8 @@ class TfqInnerProductGradOp : public tensorflow::OpKernel {
 
     // Create the output Tensor.
     const int output_dim_batch_size = context->input(0).dim_size(0);
-    const int output_dim_internal_size = context->input(3).dim_size(1);
-    const int output_dim_symbol_size = context->input(1).dim_size(0);
+    size_t output_dim_internal_size = context->input(3).dim_size(1);
+    size_t output_dim_symbol_size = context->input(1).dim_size(0);
```
MichaelBroughton, line 66: Why remove the const ?

tensorflow_quantum/core/ops/tfq_ps_weights_from_symbols_op.cc
```diff
@@ -146,7 +146,7 @@ class TfqPsWeightsFromSymbolOp : public tensorflow::OpKernel {
     auto DoWork2 = [&](int start, int end) {
       for (int i = start; i < end; i++) {
         for (int j = 0; j < n_symbols; j++) {
-          for (int k = 0; k < output_results.at(i).at(j).size(); k++) {
+          for (size_t k = 0; k < output_results.at(i).at(j).size(); k++) {
```
MichaelBroughton, line 149: Why do just k and not i, j as well ?

tensorflow_quantum/core/ops/tfq_ps_weights_from_symbols_op.cc
```diff
@@ -146,7 +146,7 @@ class TfqPsWeightsFromSymbolOp : public tensorflow::OpKernel {
...
             output_tensor(i, j, k) = output_results.at(i).at(j).at(k);
           }
           for (int k = output_results.at(i).at(j).size();
```
MichaelBroughton, line 152: Why not here ?

MichaelBroughton (approved):
Looks good. Hopefully this gets rid of some of the type warning headaches. It does seem like our use of size_t vs int in loops now will be more inconsistent overall, but if that gets us fewer warnings, I guess it's a step in the right direction.

tensorflow_quantum/core/ops/math_ops/tfq_inner_product_grad.cc
mhucka: In most of these cases, I looked to gcc's warning message and kept the changes limited to that. In these cases, the warnings were:

```
tensorflow_quantum/core/ops/math_ops/tfq_inner_product_grad.cc:184:53:
warning: comparison of integer expressions of different signedness:
'std::vector<float>::size_type' {aka 'long unsigned int'} 
and 'const int' [-Wsign-compare]
```

For whatever reason, it didn't mention `const` in relation to the size_type, so I didn't include it in the changes (and gcc doesn't warn about a missing const after the changes).

But now that you mention it, I went back and tried adding `const`, and it works, so ¯\_(ツ)_/¯. I'll update this PR to add const, and also check other cases where this might have happened.

tensorflow_quantum/core/ops/tfq_ps_weights_from_symbols_op.cc
mhucka: Ah, that's because k is the only one being compared to a `.size()` value; i and j are used as indexes.

tensorflow_quantum/core/ops/tfq_ps_weights_from_symbols_op.cc
mhucka: I agree this one looks inconsistent. It's because in this case, the value of k is being compared to `largest_single_symbol` on the next line, and that variable is an int.

The weird thing is k is initialized by setting its value to the output of `size()`, and yet the compiler isn't complaining about that part. I don't have a good explanation for that.

MichaelBroughton (approved):

PR #841, opened 2025-01-13
Label simulate_mps_test as external

mhucka:
`simulate_mps_test` takes a long time to run, and when used as part of PR checks on GitHub, it impacts developer velocity. Labeling it will let us filter it out for PR checks, and still run them as part of another longer workflow.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -68,7 +68,17 @@ jobs:
       - name: Configure CI TF
         run: echo "Y\n" | ./configure.sh
       - name: Full Library Test
-        run: ./scripts/test_all.sh
+        run: |
+          bazel test -c opt \
+          --test_timeout_filters=-eternal \
+          --test_output=errors \
+          --experimental_repo_remote_exec \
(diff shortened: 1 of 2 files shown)


MichaelBroughton (changes requested):
IIRC, these tests take a long time to compile, not nescessarily run. Because the Eigen library is a beast. Could we experiment with different flags here to cut that down before flagging it as eternal ?

mhucka:
> IIRC, these tests take a long time to compile, not nescessarily run. Because the Eigen library is a beast. Could we experiment with different flags here to cut that down before flagging it as eternal ?

It does take a long time to compile, and in an upcoming PR, compilation will be sped up via use of a bazel cache. However, `simulate_mps_test` independently takes a long time. On my cloudtop,

```
//tensorflow_quantum/core/ops/math_ops:simulate_mps_test   PASSED in 536.9s
```

That's almost 10 minutes for just that single test.

But what why don't we do this: I'll take out the eternal filter, and we can assess the run times on GitHub once all the other workflow changes are in place.

mhucka:
PR #852 is the one mentioned in my comment above.

PR #842, opened 2025-01-15
Remove lint/format checks (will put in separate files)

mhucka:
The current workflow puts all the checks in a single file. Using a single file is not necessary for some of the kinds of checks being done, and advantages can be gained by modularizing checks into separate files.

This pull request moves the lint & format checks from `ci.yaml` into new separate and modular workflows, and in addition, takes advantage of features of GitHub Actions and newer versions of the programs like `clang-format` to reduce the computation time needed by the workflows. Some of the features include:

- Trigger conditions are designed to invoke the workflows only if a pull request or push affects the relevant files. For example, the Python checks workflow is only triggered if the pull request or push affects `.py` files.

- Caching is used so that invocations usually don't need to do (for example) `pip install` every time. This makes invocation much faster.

- Colorized diff output (from `yapf` for Python and `clang-format` for C++) is displayed on the workflow summary page.

Upcoming pull requests will add additional checks workflows for other files, such as shell scripts, Markdow, and others.

### DIFF
--- a/.github/problem-matchers/pylint.json
+++ b/.github/problem-matchers/pylint.json
@@ -1,0 +1,32 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "pylint-error",
+      "severity": "error",
+      "pattern": [
+        {
+          "regexp": "^(.+):(\\d+):(\\d+):\\s(([EF]\\d{4}):\\s.+)$",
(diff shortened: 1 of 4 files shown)


mhucka:
Closing and doing it a different way in new PRs.

PR #843, opened 2025-01-17
Move CI lint/format checks into separate file

mhucka:
This is the first PR in what will be several PRs to update and improve the CI checks workflows. This PR simply moves the lint & format checks to a new, separate workflow file (`ci-file-checks.yaml`) and renames `ci.yaml` to `ci-build-checks.yaml`. There are no functional changes, apart from adjusting the conditions of the build tests in `ci-build-checks.yaml` to not condition them on passing lint/format checks.

Lint/format checks are not a strictly necessary precondition to testing builds and doing unit tests, and running them in parallel affords a couple of advantages:

- Faster overall CI execution.
- Potential for more feedback. If a lint step fails, it doesn't necessarily mean that the code won't compile, and proceeding with the build tests gives devs as much feedback as possible.

There is of course the danger that the code changes won't compile, and the resulting build will be pointless (and possibly produce confusing error messages). I think devs will be smart enough to realize that if they see _both_ lint/format errors and build errors, they should fix the former first. In addition, we can tune the conditions in the builds so that they fail early.

### DIFF
--- a/.github/workflows/ci-build-checks.yaml
+++ b/.github/workflows/ci-build-checks.yaml
@@ -2,37 +2,7 @@ name: Continuous Integration
 
 on: [pull_request]
 
-
 jobs:
-  lint:
-    name: Lint check
-    runs-on: ubuntu-20.04
-    steps:
-      - uses: actions/checkout@v1
-      - uses: actions/setup-python@v1
-        with:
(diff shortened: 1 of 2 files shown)


MichaelBroughton (approved):
LGTM

PR #844, opened 2025-01-17
Remove reference to non-existent directory

mhucka:
The `lint_all.sh` script invoked `pylint` with a couple of subdirectories, including `./examples`. However, that directory doesn't exist.

Apparently Pylint version 2.4.4 didn't complain about this, but later versions of Pylint do.


MichaelBroughton (approved):
LGTM

PR #845, opened 2025-01-17
Update version of pylint to 3.3.3

mhucka:
This updates `requirements.txt` to use Pylint version 3.3.3 and also updates `.pylintrc` such that it's compatible with 3.3.3 yet does not raise any new warnings in the code base. The intention of this PR is to make only the minimal changes needed to use pylint 3.3.3.

The `.pylintrc` started life as the config file produced by pylint 3.3.3 with the `--generate-rcfile` option. I then compared it to the previous `.pylintrc` and adjusted the settings to match, followed by a long period of test-adjust-retest:

1. run `pylint` on the existing code base
2. look at the warnings, find the relevant flags & turn them off
3. if there are more warnings, go back to step 1

This `.pylintrc` file thus has a lot of pylint checkers commented out. We should systematically re-enable the checkers and fix the warnings in the code. However, that should be left to separate PRs.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,131 +1,1014 @@
-[MASTER]
+# Summary: TFQ config file for Pylint versions 3.0 and above.
+# See https://pylint.readthedocs.io/ for info about options available.
 
+[MAIN]
+
+# Analyse import fallback blocks. This can be used to support both Python 2 and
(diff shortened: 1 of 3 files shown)


MichaelBroughton (approved):
LGTM

PR #846, opened 2025-01-17
Update yapf to version 0.43.0

mhucka:
Yapf version 0.43.0 seems to be entirely compatible with the previously-used version 0.40.2. Running it recursively on the codebase produced no output.


MichaelBroughton (approved):
LGTM

PR #847, opened 2025-01-18
Vastly overhaul the CI workflow for file checks

mhucka:
Vastly overhaul the CI workflow for file checks

This is a complete rewrite of the file checks portion of the CI workflow. It incorporates many new features and efficiency gains. It also serves as a framework for adding additional file linters/format checkers, such as for YAML, Markdown, shell scripts, and more.

Features:

- Serious use of caching of the Python installation and environment, so that repeated runs don't have to reinstall everything every time.

- Concurrency detection: if a new event such as a push to a PR branch comes in while a workflow is already executing, that workflow is cancelled and a new one restarted. (The existing workflow keeps the existing one running and starts a second one, which is almost never useful.)

- File-aware execution: the workflow tests which files have been modified in a given event, and invokes only the relevant lint/format-checker jobs. This reduces compute cycles needed and speeds up the overall execution.

- Use of GitHub [Problem Matchers](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) for better error reporting.

- Addition of a manual run interface, allowing the workflow to be invoked manually from the GitHub GUI with different values for program versions – useful for debugging, testing, exploring behaviors with different versions of linters, etc.

- Compatibility with GitHub [merge queues](https://github.blog/news-insights/product-news/github-merge-queue-is-generally-available/).

For Python, this implements the equivalent of `scripts/lint_all.sh` (which, despite its name, actually only lints Python files) and the Python-specific format checks of `scripts/format_check.sh`.

For C++, this runs clang-format with input limited to the files changed in a given PR.

PR #848, opened 2025-01-18
Add YAML file linter action

mhucka:
This extends `ci-file-checks.yaml` with a job for linting YAML files. Those files show up in the ./docs/ directory, the GitHub workflows, and some config files.

### DIFF
--- a/.github/problem-matchers/yamllint.json
+++ b/.github/problem-matchers/yamllint.json
@@ -1,0 +1,22 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "yamllint",
+      "pattern": [
+        {
+          "regexp": "^(.*\\.ya?ml)$",
+          "file": 1
+        },
+        {
+          "regexp": "^\\s{2}(\\d+):(\\d+)\\s+(error|warning)\\s+(.*?)\\s+\\((.*)\\)$",
+          "line": 1,
+          "column": 2,
+          "severity": 3,
+          "message": 4,
+          "code": 5,
+          "loop": true
+        }
+      ]
+    }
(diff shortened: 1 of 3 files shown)


MichaelBroughton (commented):
Some of these look significantly more complicated (in terms of bash commands) relative to existing CI jobs. How does a user run this locally ?

mhucka:
> Some of these look significantly more complicated (in terms of bash commands) relative to existing CI jobs. How does a user run this locally ?

First, a clarification: this PR only adds yaml linting, and it's meant to be merged after PR #847 is merged. The diff is unfortunately much larger than it would be otherwise, because it's a PR done off a branch that contains the changes of #847 (because it needs the scaffolding of that other PR).

So, let's consider PR #847 in order to answer the question. With respect to running this locally: the workflow looks very complicated only because of the tests and caching done to make it efficient, and for reporting results in the GitHub interface. The actual file check CI operations done by PR #847 are just this:

- `clang-format --verbose --style google --dry-run` on the changed files
- `pylint` on the changed files
- `yapf --parallel --diff --style=google` on the changed files

Those 3 things are what the two scripts `scripts/lint_all.sh` and `scripts/format_check.sh` ultimately do, so there is no change to how users would run those locally.

For the addition in this PR (#848), it adds `yamllint` as an additional job in the workflow (conditioned on whether any YAML files were changed in the PR). That doesn't have a `script/foo` equivalent right now. Should we add `yamllint` there somewhere? Perhaps in script/lint_all.sh`?

MichaelBroughton (approved):
LGTM

Issue #849, opened 2025-01-19
Incorrect parameterization args to test in noisy_controlled_qpc_test.py


mhucka:
There's something odd happening with the parameter value in this test in TFQ:  https://github.com/tensorflow/quantum/blob/452a2127c16de6d65212299f811f34a9c66b5c3e/tensorflow_quantum/python/layers/high_level/noisy_controlled_pqc_test.py#L149

A dict is being created with the same key, but different values. In Python, that means the last assignment wins; it won't result in 2 dict entries. Therefore, that part of the code is probably wrong.

mhucka:
Completed by #850

PR #850, opened 2025-01-19
Fix issue #849: incorrect argument passed to test function

mhucka:
This fixes the incorrect `dict` being passed as an argument to the test function `test_controlled_pqc_simple_learn`, reported in issue #849.


MichaelBroughton (approved):
LGTM

PR #852, opened 2025-01-20
Vastly overhaul the CI workflow for build checks

mhucka:
This workflow compiles TFQ and runs test cases to verify things work after file changes are made by a PR or other event. It is a complete rewrite of the previous TFQ build & test checks CI workflow, and incorporates many new features and efficiency gains:

- Serious use of caching of the Python installation and environment, as well as the Bazel environment, so that repeated runs don't have to reinstall everything every time.

- File-aware execution: the workflow tests whether source files and build configuration files have been modified in a given event, and stops early if the event didn't affect such files.

- Compatibility with GitHub [merge queues](https://github.blog/news-insights/product-news/github-merge-queue-is-generally-available/).

- Concurrency detection: if a new event such as a push to a PR branch comes in while a workflow is already executing, that workflow is cancelled and a new one restarted. (The existing workflow keeps the existing one running and starts a second one, which is almost never useful.)

- Addition of a manual run interface, allowing the workflow to be invoked manually from the GitHub GUI with different values for program versions – useful for debugging, testing, exploring behaviors with different versions of linters, etc.

### DIFF
--- a/.github/workflows/ci-build-checks.yaml
+++ b/.github/workflows/ci-build-checks.yaml
@@ -1,84 +1,508 @@
-name: Continuous Integration
+# Summary: TFQ continuous integration workflow for building & testing TFQ.
+#
+# This workflow compiles TFQ and runs test cases to verify everything works.
+# It triggers on certain events such as pull requests and merge-queue merges,
+# tries to be as efficient as possible by caching the Python environment and
+# Bazel artifacts, and can be invoked manually via the "Run workflow" button at
+# https://github.com/tensorflow/quantum/actions/workflows/ci-build-checks.yaml
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-on: [pull_request]
+name: CI build checks
+run-name: Continuous integration build & test
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+    branches:
+      - main
+
+  merge_group:
+    types:
+      - checks_requested
+
+  push:
+    branches:
+      - main
+
+  # Allow manual invocation, with options that can be useful for debugging.
+  workflow_dispatch:
+    inputs:
+      sha:
+        description: "SHA of commit to run against:"
+        type: string
+        required: true
+
+      py_version:
+        description: "Python version:"
+        type: string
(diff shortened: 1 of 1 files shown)


mhucka:
In reference to a previous PR proposing to filter test execution to exclude tests tagged as "eternal", here is some timing information.

1. The basic workflow, with no caching or other enhancements, takes about 38 minutes to run if Bazel test timeouts are set long enough to allow all tests to complete. Here's a screenshot using the new "ci-nightly-check.yaml" workflow (it's in a separate PR), whose purpose is to run once a day to build and do all the tests without doing anything clever:

<div align="center">

</div><br>

2. With the new build & test workflow (the present PR), and again setting timeouts as long as possible (not the default, rather using `--test_timeout=6000`) so that we can see how long the tests themselves take to run to completion, the overall time is like this:
image
The new workflow takes only ~2 minutes to build TFQ. The TFQ tests take around 17 minutes and the tutorial tests take 10-11 minutes. Examining the logs shows that the longest times are for the gradient and sometimes some other tests. Here are the 3 longest in an example run:
    ```
    //tensorflow_quantum/core/ops:circuit_execution_ops_test   PASSED in 385.1s
    //tensorflow_quantum/python/differentiators:gradient_test  PASSED in 575.9s
    //tensorflow_quantum/core/ops:tfq_unitary_op_test          PASSED in 306.6s
    ```
    The `cicuit_execution_ops_test` and `gradient_test` are already both tagged as "eternal". Interestingly, `tfq_unitary_op_test` is not so tagged. With respect to the discussion in [another PR](https://github.com/tensorflow/quantum/pull/841#pullrequestreview-2560239960), it's interesting that in this example, `simulate_mps_test` is not one that showed up this time. It has definitely shown up before with high times, so I think we're just seeing some expected stochastic variability going on.

3. Finally, using the default Bazel test timeouts, here's the timing:
image

These are shorter because they're hitting timeouts and getting cut short.

Conclusions:

- This new build & test CI workflow significantly reduces the overall time to do PR checks.
- The run times are now constrained by the time it takes to run the tests.
- The tutorials take a long time to test. Even if we manage to reduce the time to do the Bazel tests, the overall time won't go less than about 13-15 minutes unless we reduce the 10-11 minute time for the tutorial tests. Finding ways of shortening the tutorial tests is something worth exploring.

MichaelBroughton (approved):
LGTM

PR #853, opened 2025-01-20
Add workflow for nightly full test

mhucka:
This workflow does a plain TFQ build & test. It's scheduled to run every night. Its purpose is to guard against failures that might be missed when skipping the long-running tests or using caching to speed up the CI runs. Unlike the CI checks invoked on PRs and similar events, this workflow builds everything without caching and runs the full test suite, including "eternal" tests that take a long time to run.


MichaelBroughton (approved):
LGTM

PR #854, opened 2025-01-20
Add new Trove classifier term for quantum computing

mhucka:
PyPI now supports a classifier term for quantum computing.


MichaelBroughton (approved):
LGTM

PR #858, opened 2025-01-24
Add YAML file check

mhucka:
This adds a job to run `yamllint` on `*.{yaml,yml}` files in the repository. YAML files are currently found not only in the GitHub workflows directory; they're also in `docs/`, and we can imagine more YAML-based config files being added in the future.


MichaelBroughton (approved):
LGTM

PR #864, opened 2025-02-27
Add .editorconfig file

mhucka:
This adds settings based on some common-sense values and the TensorFlow Quantum project's current conventions, such as line length and use of spaces instead of tabs. The settings match the corresponding ones in `.pylintrc`.


MichaelBroughton (approved):
LGTM

PR #878, opened 2025-04-14
Constrain numpy version to be less than 2.x

mhucka:
Although the latest release of Cirq is compatible with NumPy 2, other
dependencies for TFQ are not, and this leads to failures at run-time.
For now, tell `pip` not to install NumPy 2.

### DIFF
--- a/.github/workflows/ci-nightly-cirq-test.yaml
+++ b/.github/workflows/ci-nightly-cirq-test.yaml
@@ -92,6 +92,8 @@ jobs:
 
       - name: Install the nightly build version of Cirq
         run: |
+          echo 'numpy<2.0.0' > constraint.txt
+          export PIP_CONSTRAINT=constraint.txt
           pip install -U cirq --pre
 
       - name: Configure Bazel options


mhucka:
> Does this cause a breakage in the pip package too ? If so we might want to put this constraint there as well ?

It seems like that would make sense, but unfortunately, it doesn't end up working. It's because the sequence of commands is, first,

```bash
pip install -r requirements.txt
```

then followed separately by 

```bash
pip install -U cirq --pre
```

and while we could constrain the version of NumPy in the requirements file to be < 2.0, the second command (installing Cirq) will cause NumPy 2 to be brought in. Pip doesn't remember the constraints between invocations.

I think what needs to happen is to make TFQ work with Cirq 1.5, at which point, the version of Cirq defined in requirements.txt can be set to 1.5 (instead of 1.3 as it is now), and _then_ the version of NumPy in requirements.txt can be limited to < 2.0. Having the requirements and packages be installed in one go will make pip satisfy the simultaneous constraints.

For now, changing the compatibility workflow to use the pip constraints feature will let the CI checks pass.

MichaelBroughton (approved):
SGTM

PR #883, opened 2025-08-28
Update README file overall

mhucka:
This updates the info, look & feel, adds some more feature points, adds Google disclaimer, and does misc tweaks.


mhucka:
> These changes look fine. If we want a transparent version of the logo, can we just have one and give it the same name as the original logo file with _transparent, so people know the difference. otherwise LGTM.

That's a fine idea – better than what I did. I'll change the file, delete the extra png, and update the PR.

Thanks.

PR #901, opened 2025-09-14
Add Scorecard and OSV security scanning workflows 

mhucka:
Google's OSS guidance recommend the use of security scanners to detect vulnerabilities in code. This PR adds Scorecard and the Open-Source Vulnerabilities (OSV) scanners. Both of these are being used in our Quantumlib repos.

The configurations here will run the scanners on these events:
* pull requests
* merge queue merges
* manual invocation
* weekly on Saturdays


mhucka:
Will redo.

PR #913, opened 2025-11-07
TF2.16 support: hermetic python shim, TF 2.16.2 pin, legacy Keras path, setup/packaging fixes

psamanoelton:
Summary
- Enable building TFQ against TensorFlow 2.16.2 with Bazel 6.5.x
- Keep Keras 2 (`tf-keras`) path via `TF_USE_LEGACY_KERAS=1` to avoid Keras 3 breakages
- Allow choosing Python interpreter, whether from system, venv, or conda
- Make wheel-build script honor chosen Python and include setuptools
- Define essential requirements in `requirements.in` and use `pip-compile` to produce `requirements.txt`

Key changes
- WORKSPACE:
   - bump `@org_tensorflow` to 2.16.2
   - use newer `rules_python` Bazel rule set
   - use `pip_parse` and `install_deps` from `rules_python` to read `requirements.txt` and install dependencies
   - add tiny `@python` local_repository shim to set Python interpreter path
- `third_party/python_legacy/defs.bzl`:
   - export interpreter as a *string* (what TF’s local_config_python expects)
- `configure.sh`:
   - new `--python=` flag
   - write `.tf_configure.bazelrc` + `.bazelrc` + the Python shim
   - make `.bazelrc` also import `.bazelrc.user` if it exists
- `.bazelrc`:
   - ensure `TF_USE_LEGACY_KERAS=1` for build/test
   - keep existing warnings/rpath tweaks
- `release/build_pip_package.sh`:
   - invoke `${PYTHON_BIN_PATH:-python3}` and ensure setuptools present
- `setup.py`:
   - (optionally) relax deps to run on TF 2.16 (`sympy==1.14`)
   - extras pin `tensorflow>=2.16,<2.17`
   - alternative name `and-tensorflow` for extras lets users run `pip install tensorflow-quantum[and-tensorflow]` for a more intuitive installation command
- `requirements.in`:
   - set of requirements needed for building TFQ and running test scripts
   - `requirements.txt` is produced by running `pip-compile --allow-unsafe`
 
Testing
- Linux/Ubuntu 22.04 (Docker), Python 3.11, Bazelisk → Bazel 6.5.0
- `bazel build ... release:build_pip_package` succeeds
- Unit tests (`scripts/test_all.sh`) pass
- Wheel installs and imports with TF 2.16.2 + tf-keras 2.16.0, Cirq 1.3.0, NumPy 1.26.4

Miscellaneous
- Run changed Python files through formatter

Notes
- This keeps TFQ on Keras 2 path; Keras 3 migration can be handled in a a future PR.
- This changes will help on the incoming upgrades to 2.20.

### DIFF
--- a/scripts/benchmark_all.sh
+++ b/scripts/benchmark_all.sh
@@ -1,12 +1,12 @@
 #!/bin/bash
 # Copyright 2020 The TensorFlow Quantum Authors. All Rights Reserved.
-#
+# 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-#
+# 
 #     http://www.apache.org/licenses/LICENSE-2.0
-#
+# 
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -14,7 +14,7 @@
 # limitations under the License.
 # ==============================================================================
 echo "Testing benchmarks.";
-test_outputs=$(bazel test -c opt --cxxopt="-D_GLIBCXX_USE_CXX11_ABI=1" --cxxopt="-msse2" --cxxopt="-msse3" --cxxopt="-msse4" --test_output=errors "$(bazel query //benchmarks/...)")
+test_outputs=$(bazel test -c opt --cxxopt="-D_GLIBCXX_USE_CXX11_ABI=1" --cxxopt="-msse2" --cxxopt="-msse3" --cxxopt="-msse4" --test_output=errors $(bazel query //benchmarks/...))
 exit_code=$?
 
 if [ "$exit_code" == "0" ]; then
--- a/scripts/build_pip_package_test.sh
+++ b/scripts/build_pip_package_test.sh
@@ -1,12 +1,12 @@
 #!/bin/bash
 # Copyright 2020 The TensorFlow Quantum Authors. All Rights Reserved.
-#
+# 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-#
+# 
 #     http://www.apache.org/licenses/LICENSE-2.0
-#
+# 
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -17,7 +17,7 @@
 pip install -r requirements.txt
 
 # cd tensorflow_quantum
-printf "y\n" | ./configure.sh
+echo "Y\n" | ./configure.sh
 
 bazel build -c opt --cxxopt="-D_GLIBCXX_USE_CXX11_ABI=1" --cxxopt="-msse2" --cxxopt="-msse3" --cxxopt="-msse4" release:build_pip_package
 rm /tmp/tensorflow_quantum/* || echo ok


release/build_pip_package.sh
```diff
@@ -16,44 +16,58 @@
...
 
-function main() {
-  DEST=${1}
-  EXTRA_FLAGS=${2}
+main() {
+  DEST="$1"
+  EXTRA_FLAGS="$2"
 
-  if [[ -z ${DEST} ]]; then
+  if [[ -z "$DEST" ]]; then
```
mhucka: Here and elsewhere that variables are referenced, could we keep the recommended Google style for shell scripts? In particular, for this case the section on [variable expansion](https://google.github.io/styleguide/shellguide.html#variable-expansion) is relevant, and the preference is for brace-delimiting variables: "${DEST}".

release/build_pip_package.sh
```diff
@@ -16,44 +16,58 @@
 set -e
 set -x
 
+# Pick the Python that TFQ/TensorFlow used during configure/build.
+# Order: explicit env -> 3.11 -> python3
+PY="${PYTHON_BIN_PATH:-}"
+if [[ -z "$PY" ]]; then
+  if command -v python3.11 >/dev/null 2>&1; then
```
mhucka: We should probably not force python3.11 unconditionally in the build scripts. It would probably be best to use `python3` here, and leave it to the calling environment to make sure the version of Python meets a minimum.

(And now that this comes up, I can't find a place in the code where a minimum version is set. It would be at least in the `setup.py` file, but it doesn't have anything. I'll open an issue.)

release/build_pip_package.sh
```diff
@@ -16,44 +16,58 @@
...
+  fi
+fi
+echo "Using Python: $PY"
+
+# Ensure packaging tools are present in THIS interpreter
+"$PY" - <<'PY' || { "$PY" -m pip install --upgrade pip setuptools wheel; }
+import importlib
+for m in ["setuptools","wheel"]:
+    importlib.import_module(m)
+PY
```
mhucka: This is an unconventional way of testing that certain packages have been installed. Is it not possible to use `pip show` and test its return value? For example, something like

```shell
if ! "$PY" -m pip show -q setuptools wheel > /dev/null 2>&1; then
  "$PY" -m pip install --upgrade pip setuptools wheel
fi
```

configure.sh
```diff
@@ -13,167 +13,208 @@
...
+  read -p "Build against TensorFlow CPU pip package? [Y/n] " INPUT || true
+  case "${INPUT:-Y}" in
+    [Yy]* ) echo "CPU build selected."; TF_NEED_CUDA=0;;
+    [Nn]* ) echo "GPU build selected."; TF_NEED_CUDA=1;;
+    * ) echo "Please answer Y or n.";;
   esac
 done
 
+# For TF >= 2.1 this value isn’t actually consulted by TFQ,
+# but we keep a compatible prompt/flag.
```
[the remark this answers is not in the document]
samanoelton: I kept this from the previous code, I keep this because I am not sure if it might be needed for any other package or dependency.

release/build_pip_package.sh
```diff
@@ -16,44 +16,66 @@
...
+  EXTRA_FLAGS="$2"
 
-  if [[ -z ${DEST} ]]; then
+  if [[ -z "${DEST}" ]]; then
     echo "No destination directory provided."
     exit 1
   fi
 
-  mkdir -p ${DEST}
+mkdir -p "${DEST}"
```
mhucka: Trivial nit: something happened to the intendentation on this line.

```suggestion
  mkdir -p "${DEST}"
```

release/setup.py
```diff
@@ -12,105 +12,106 @@
...
-    long_description='\n'.join(DOCLINES[2:]),
-    author='Google Inc.',
-    author_email='no-reply@google.com',
-    url='https://github.com/tensorflow/quantum/',
+    name=PROJECT_NAME,
+    version=BUILD_VERSION,
+    description="Library for hybrid quantum-classical machine learning.",
+    long_description="\n".join(DOCLINES[2:]),
+    author="Google Inc.",
+    author_email="no-reply@google.com",
```
mhucka: Since this file is getting changes for other reasons, it may as well also include an update to these values. The general approach we've taken in other major QAI projects is to use the following pattern, so I think it's worth including in this PR so that another PR isn't needed just for this.

```suggestion
    author="The TensorFlow Quantum Authors",
    author_email="tensorflow-quantum-team@google.com",
```

configure.sh
```diff
@@ -13,167 +13,235 @@
...
-function write_action_env_to_bazelrc() {
-  write_to_bazelrc "build --action_env $1=\"$2\""
+# --- helpers ---------------------------------------------------------------
+write_bazelrc() {
+  echo "${1}" >> .bazelrc
 }
 
-function write_linkopt_dir_to_bazelrc() {
-  write_to_bazelrc "build --linkopt -Wl,-rpath,$1" >> .bazelrc
+write_tf_rc() {
```
mhucka:
```suggestion
function write_tf_rc() {
```

configure.sh
```diff
@@ -13,167 +13,235 @@
...
+  echo "${1}" >> .tf_configure.bazelrc
 }
 
-
-function is_linux() {
-  [[ "${PLATFORM}" == "linux" ]]
+die() {
```
mhucka:
```suggestion
function die() {
```

configure.sh
```diff
@@ -13,167 +13,235 @@
...
 
-function is_windows() {
-  # On windows, the shell script is actually running in msys
+is_windows() {
   [[ "${PLATFORM}" =~ msys_nt*|mingw*|cygwin*|uwin* ]]
 }
 
-function is_ppc64le() {
-  [[ "$(uname -m)" == "ppc64le" ]]
+write_legacy_python_repo() {
```
mhucka:
```suggestion
function write_legacy_python_repo() {
```

configure.sh
```diff
@@ -13,167 +13,235 @@
...
+# Normalize to an absolute path (readlink -f is GNU; fall back to python)
+if command -v readlink >/dev/null 2>&1; then
+  PY_ABS="$(readlink -f "${PY}" 2>/dev/null || true)"
+fi
+if [[ -z "${PY_ABS:-}" ]]; then
+  PY_ABS="$("${PY}" - <<'PY'
+import os,sys
+print(os.path.abspath(sys.executable))
+PY
+)"
```
mhucka: Although this works, IMHO this construct is fragile because of the complicated syntax. (At first, I thought the `)"` by itself at the end was a typo.) Would the following not work?

```suggestion
  PY_ABS="$("${PY}" -c 'import os,sys; print(os.path.abspath(sys.executable))')"
```

configure.sh
```diff
@@ -13,167 +13,235 @@
...
+LIBNAME="$("$PYTHON_BIN_PATH" - <<'PY'
+import os, glob, tensorflow.sysconfig as sc
+p = sc.get_lib()
+d = p if os.path.isdir(p) else os.path.dirname(p)
+cands = glob.glob(os.path.join(d,'libtensorflow_framework.so*')) \
+     or glob.glob(os.path.join(d,'libtensorflow.so*')) \
+     or glob.glob(os.path.join(d,'_pywrap_tensorflow_internal.*'))
+print(os.path.basename(cands[0]) if cands else 'libtensorflow_framework.so.2')
+PY
+)"
```
mhucka: This is going to be slow because it imports tensorflow 4 times. It's true that the imports after the first time will be faster thanks to caching by the operating system, but still, it feels like there must be a more efficient way than running Python 4 times. Also, there is a little bit of duplication of code.

What about the following?
```suggestion
tf_output=$("${PYTHON_BIN_PATH}" - <<'PY'
import sys
import os
import glob

try:
    import tensorflow as tf
    import tensorflow.sysconfig as sc
except ImportError:
    sys.exit(1)

print(sc.get_include())

lib_path = sc.get_lib()
lib_dir = lib_path if os.path.isdir(lib_path) else os.path.dirname(lib_path)
print(lib_dir)

cands = (glob.glob(os.path.join(lib_dir, 'libtensorflow_framework.so*')) or
         glob.glob(os.path.join(lib_dir, 'libtensorflow.so*')) or
         glob.glob(os.path.join(lib_dir, '_pywrap_tensorflow_internal.*')))
print(os.path.basename(cands[0]) if cands else 'libtensorflow_framework.so.2')
PY
)

if [[ $? -ne 0 ]]; then
  echo "ERROR: tensorflow not importable by Python (${PYTHON_BIN_PATH})" >&2
  exit 1
fi

{
  read -r HDR
  read -r LIBDIR
  read -r LIBNAME
} <<< "${tf_output}"
```

WORKSPACE
```diff
@@ -64,7 +65,6 @@ tf_workspace1()
 
 load("@org_tensorflow//tensorflow:workspace0.bzl", "tf_workspace0")
 
-tf_workspace0()
```
mhucka: Is the removal of this call to `tf_workspace0()` intentional?

PR #924, opened 2025-12-09
Add more support materials for contributors

mhucka:
This adds information about project conventions and workflows to `CONTRIBUTING.md`, and in support of that, also adds configuration files for more linters and formatters.

The biggest change is to `CONTRIBUTING.md`, which is now greatly expanded. It includes more "getting started" type guidance, plus developer processes, and coding conventions. Where possible, this references other documentation in TensorFlow or TensorFlow Quantum.

The remaining changes are to help developers get the right settings for some commonly-used programs, add GitHub problem matchers for more of the programs used in CI, and add a couple of files that are standard these days:

- Add configuration files for some linters and formatters relevant to this project. Not only can users run, e.g., `clang-format /path/to/some/file` and it will use the correct format without the user having to remember to add a style parameter; program config files are recognized by some editing tools to help developers get immediate feedback while they're editing. (E.g., Emacs has such tooling.)

- Add more problem matchers to help surface errors in CI by taking advantage of the GitHub [problem matchers](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) feature

- Add a code of conduct and a standard project `SUPPORT.md` file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -173,7 +173,7 @@ section, you should be able to build TFQ with these steps:
 2.  Build a pip package for TFQ:
 
     ```shell
-    bazel build -c opt release:build_pip_package
+    bazel build -c opt --cxxopt="-O3" --cxxopt="-march=native" release:build_pip_package
     bazel-bin/release/build_pip_package /tmp/tfquantum/
     ```
 


MichaelBroughton (changes requested):
Thanks for working on this. I don't think we can merge this PR as is. There are too many changes at once in a single PR. I see the git-blame-ignore-revs stuff is in here again, which I don't think we want. It looks like there is a whole bunch of other linting and styling changes going on here that are hard to track where they come from and how it all fits together on top of all of the changes to `.md` files. Could you please split this into smaller focused PRs ?

Would also recommend reading go/prefer-small-cls for some general guidelines when splitting things up.

PR #925, opened 2025-12-09
Add new scripts for building and cleaning wheels

mhucka:
These new scripts in `release/` encapsulate instructions from Michael Broughton about how to build new distributions of TensorFlow Quantum. The file `release/README.md` describes how to use them. There are unit tests as well; they can be invoked using bazel (and the process is also described in the README file).

There are a couple of tweaks to `release/build_pip_package.sh` to take a debug tracing argument and make the script work inside the Docker containers run by the build script.


MichaelBroughton (commented):
Thanks for taking a stab at writing down the instructions. A few minor nits in the bash files themselves and then consider removing the python tests. After that LGTM.

PR #926, opened 2025-12-09
Formatting only: run code through scripts/format_all.sh

mhucka:
Some trivial deviations from `clang-format` in a small number of files seem to have crept in over time. This meant that running the format check scripts in `scripts/` could report changes that are not a contributors' actual changes. To prevent possible confusion to developers and maintainers, this commit:

1. adds the results of running `./scripts/format_all.sh`
2. adds a `.git-blame-ignore-revs` file
3. adds intructions to the contributor instructions to run `git config blame.ignoreRevsFile` on the file

⚠️ Maintainer notice ⚠️: this PR needs to be merged using the "**Create a merge commit**" strategy. Using "Squash and Merge" or "Rebase and Merge" will change the commit hash and render the entry in `.git-blame-ignore-revs` ineffective.


MichaelBroughton (commented):
Thanks for the formatting updates! Happy to merge those, but I don't think we should take on the extra maintenance burden of the the blame-ignore-revs here since it looks like it adds complexity for how we need to merge PRs that isn't worth putting on people anytime we have a formatting change that they want to add to this list. This is a small repo and we need to keep these sorts of things as small as possible.

PR #933, opened 2025-12-10
Use larger runners for nightly jobs

mhucka:
The nightly jobs are failing due to running out of disk space. (It's the same problem we had recently with the CI checks workflow.) This PR updates the jobs to use the larger self-hosted runners from Google ML Velocity. 

In addition, this PR makes some simplifications to the workflows to get rid of some of the overelaborate things I did before, and make minor other changes to keep with best practices such as putting timeouts on jobs.

### DIFF
--- a/.github/workflows/ci-nightly-build-test.yaml
+++ b/.github/workflows/ci-nightly-build-test.yaml
@@ -80,7 +80,7 @@ jobs:
       have-changes: ${{steps.commits.outputs.count > 0}}
     steps:
       - name: Check out a sparse copy of the git repo for TFQ
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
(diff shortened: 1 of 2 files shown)


MichaelBroughton (changes requested):
This seems like it's combining a lot of changes all at once that may interact in complex ways. Can we split this into two PRs ? The first being just the runner upgrade, very simple and short. Verify that this is working well and doesn't break. Then we can move onto doing all of the other stuff:
```
In addition, this PR makes some simplifications to the workflows to get rid of some of the overelaborate things I did before, and make minor other changes to keep with best practices such as putting timeouts on jobs.
```

mhucka:
> This seems like it's combining a lot of changes all at once that may interact in complex ways. Can we split this into two PRs ? The first being just the runner upgrade, very simple and short. Verify that this is working well and doesn't break. Then we can move onto doing all of the other stuff:

Yes, that's reasonable.

PR #934, opened 2025-12-11
Further constrain dependency versions in setup.py and requirements.txt

mhucka:
This PR addresses three needs with respect to dependency management and updates both `setup.py` and `requirements.txt` to add some constraints on a few dependencies in order to make installations trouble-free.

* In some environments, we need to constrain the versions of certain Python packages required to pip-install TFQ in order to avoid pip dependency resolution errors. The two problems that I'm seeing on Linux and Colab are (a) pip may try to install a version of the dependency that requires a higher version of Python than what the user is running, an (b) pip tries to install or build NumPy 2 because some dependency version constraints NumPy to be >2. To deal with this, this PR adds a few additional version constraints to `setup.py` for packages that end up needing to be installed when installing TFQ. These additions to `setup.py` are the minimum version constraints that produced error-free installations in my testing.

* Similar dependency version constraints need to be specified for developers installing dependencies using `requirements.txt`. However, a better approach for `requirements.txt` is to use a minimal `requirements.in` file in combination with `pip-compile` to write the final `requirements.txt`. The resulting `requirements.txt` file leads to more predictable build & test environments.

* A shortcoming of generating `requirements.txt` with version pins is that the versions of some transitive dependencies may need to be updated to address security releases. Simply re-running `pip-compile` to produce a new `requirements.txt` may not update the versions because the dependencies that bring in the transitive dependencies have not themselves changed. The recommended practice for this situation is to put the version constraints into a separate constraint file and pass it as an argument to `pip-compile`. (`pip-compile` has explicit support for this.) Some additions to the script `scripts/generate_requirements.sh` encapsulate and document this process.


samanoelton (approved):
LGTM

PR #945, opened 2025-12-23
Update requirements and scripts/generate_requirements.sh

mhucka:
This PR updates the dependencies in `requirements.in` and `setup.py` as follows:

* Drop support for Python 3.9 (which was end-of-lifed last year)
* Update the requirements to try to reduce them to the minimum

This PR also updates `scripts/generate_requirements.sh` to remove some no-longer-needed post-processing of the `requirements.txt` file. The modified script still needs to have one special case, to avoid pinning the version of PyYAML. PyYAML is a transitive dependency that, in older Pythons, was installed via distutils and not pip. Pinning the version (as pip-compile would normally do) causes pip to try to uninstall any existing versions of pyyaml in the environment, and fail, because pip didn't install it in the first place. The most robust solution I've found is to avoid pinning the version. Previously this was accomplished in `scripts/generate_requirements.sh` by editing the output using `sed` (ugh – my bad). This update uses pip-compile's feature for "unsafe" packages to simply leave pyyaml out of `requirements.txt`.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -68,12 +68,12 @@ cc_library(
 
 http_archive(
     name = "qsim",
-    sha256 = "b9c1eba09a885a938b5e73dfc2e02f5231cf3b01d899415caa24769346a731d5",
+    sha256 = "720eeb97298819e00bbb218b8b58fcebbbc1e1708233598fdffeef0b97339617",
     # patches = [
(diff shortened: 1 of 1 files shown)


MichaelBroughton (approved):
LGTM

PR #948, opened 2025-12-26
Add new scripts for creating Docker containers to enable better testing of TFQ builds

mhucka:
This as a subdirectory `release/docker/` containing a `Dockerfile` and a script to create Docker images. The images are basic Ubuntu Linux 22.04 and 24.04, with a version of Python preinstalled.  Running the script `create_docker_images.sh` in that directory creates a total of eight images, with names like `ubuntu22-cp39`, `ubuntu22-cp310`, etc.

These are useful for testing TensorFlow Quantum builds in more isolated environments with different versions of Python. More information can be found in the README file `release/docker/README.md`.

The setup is not fancy, but it is useful when (e.g.) trying different combinations of dependency versions with different Python versions.


mhucka:
Closing because this branch was accidentally pushed to the origin instead of my fork.

PR #957, opened 2025-12-29
Fix #955: replace use of pkg_resources

### LINKED ISSUE #955: pkg_resources is deprecated; tutorials should be updated to use something else
All the tutorials for TFQ currently have the following code:

```python
# Update package resources to account for version changes.
import importlib, pkg_resources
importlib.reload(pkg_resources)
```

This produces the following warnings in all the tutorials:

<div align="center">
  Image
</div>

This code should be replaced with more modern equivalents that don't produce a warning.

mhucka:
All the tutorials for TFQ currently have the following code:

```python
# Update package resources to account for version changes.
import importlib, pkg_resources
importlib.reload(pkg_resources)
```

This produces warnings when running the tutorials because `pkg_resources` is deprecated. This PR replaces the code with alternatives that are considered the modern equivalents.

### DIFF
--- a/docs/tutorials/barren_plateaus.ipynb
+++ b/docs/tutorials/barren_plateaus.ipynb
@@ -110,9 +110,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Update package resources to account for version changes.\n",
-    "import importlib, pkg_resources\n",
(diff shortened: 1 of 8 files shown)


mhucka:
We're going to remove this part altogether.

PR #969, opened 2026-01-12
Update `.editorconfig`

mhucka:
This makes a couple of corrections to `.editorconfig` and reformats the comments at the top to be less superflous and more in line with the style in other TFQ files.

* Add a missing entry for Bazel `WORKSPACE` files
* Correct the indentation for shell scripts to be 2 instead of 4 (so that it matches what TensorFlow uses
* Add missing curly braces in the entry for YAML files


MichaelBroughton (approved):
LGTM

PR #970, opened 2026-01-12
Use https://github.com/bazel-contrib for rules_python

mhucka:
It looks like rules_python was moved to https://github.com/bazel-contrib from https://github.com/bazelbuild at some point in the recent past. I get occasional failures when using the latter (maybe there's a redirect?) and more consistent success with the former.


MichaelBroughton (approved):
LGTM

PR #1035, opened 2026-04-06
Fixes to benchmark test code due to missing updates in dependency methods

kenya-sk:
This PR fixes benchmark breakages in `benchmarks/scripts` and restores Bazel test coverage for the benchmark targets.

**Changes**

- Fixed missing Bazel deps in `benchmarks/scripts/BUILD` and made `random_clifford_circuit` importable by switching to py_library in `benchmarks/scripts/models/BUILD`.
- Fixed TFQ gradient benchmark input handling in `benchmark_op_gradients.py` by passing symbol_names as a tensor.
- Replaced removed Cirq API usage in `benchmark_random_circuit.py` with an in-file implementation based on ReCirq’s `google_v2_beyond_classical.py` logic (with source attribution), while preserving existing benchmark behavior (fixed-depth output).

**Note**
Direct ReCirq adoption was intentionally deferred in this PR because it introduces a broader dependency migration than needed for a benchmark-fix change. In this repo, adding ReCirq as a required dependency impacts Python lockfile resolution and Bazel external dependency wiring, and we hit dependency-resolution conflicts while trying to make it mandatory. To keep this PR focused, low-risk, and test-stable, we inlined the minimal circuit-generation logic with source attribution, and left full ReCirq integration as follow-up work once dependency constraints are aligned.

**Results**
All benchmark tests have passed.

```
$ bash scripts/test_benchmarks.sh 
Testing all Benchmarks.
Loading: 0 packages loaded
Loading: 0 packages loaded
INFO: Analyzed 6 targets (21 packages loaded, 225 targets configured).
INFO: Found 2 targets and 4 test targets...
INFO: Elapsed time: 5.054s, Critical Path: 4.64s
INFO: 2 processes: 1 internal, 1 local.
INFO: Build completed successfully, 2 total actions
//benchmarks/scripts:benchmark_clifford_circuit                 (cached) PASSED in 6.9s
//benchmarks/scripts:benchmark_op_gradients                     (cached) PASSED in 8.9s
//benchmarks/scripts:benchmark_util_test                        (cached) PASSED in 4.3s
//benchmarks/scripts:benchmark_random_circuit                            PASSED in 4.6s

Executed 1 out of 4 tests: 4 tests pass.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
Testing Complete!
```


benchmarks/scripts/benchmark_random_circuit.py
```diff
@@ -32,14 +34,96 @@
 TEST_PARAMS_1 = flags.test_flags(n_rows=3, n_cols=5, n_moments=5)
 TEST_PARAMS_2 = flags.test_flags(n_rows=4, n_cols=4, n_moments=20)
 
+T = TypeVar('T')
+
+
+def _choice(rand_gen: float, sequence: Sequence[T]) -> T:
+    """Choose a pseudo-random element from a non-empty sequence."""
+    return sequence[int(rand_gen * len(sequence))]
```
mhucka: Can you check that the Python function `random.choice(sequence)` does not already do what this `_choice()` does?

mhucka:
@kenya-sk I made one suggestion about the `_choice` function; in addition, the comments from Gemini Code Assist seem reasonable too. If you could update the branch and consider the suggestions, then I'll give it another quick review and I think it will be good to go.

benchmarks/scripts/benchmark_random_circuit.py
kenya-sk: I checked `random.choice(sequence)`, and it is not fully equivalent here. `_choice()` is intentionally aligned with the [original ReCirq implementation](https://github.com/quantumlib/ReCirq/blob/9260c35feb182d430152a513dec90cdd5203cdcd/recirq/beyond_classical/google_v2_beyond_classical.py#L234): it uses the `random.Random(seed).random` float stream directly to preserve the same seeded circuit generation behavior. 
`random.choice()` uses a different internal mechanism (`_randbelow`/`getrandbits`), which can change the selected gates even with the same seed. 

To preserve reproducibility relative to the source implementation, I’ll keep `_choice()` and add a comment explaining why it is used instead of `random.choice()`.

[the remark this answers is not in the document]
kenya-sk: I updated `_choice()` to match the [original ReCirq pattern](https://github.com/quantumlib/ReCirq/blob/9260c35feb182d430152a513dec90cdd5203cdcd/recirq/beyond_classical/google_v2_beyond_classical.py#L234) by passing the RNG callable itself, which makes `rand_gen` accurate here and keeps the seeded circuit-generation behavior explicit.

benchmarks/scripts/benchmark_random_circuit.py
```diff
@@ -32,14 +34,96 @@
...
+        ]:
+            if q2 in qubits and ((q.row + q.col + offset) % 2 == 0):
+                yield cirq.CZ(q, q2)
+
+
+def _add_cz_layer(layer_index: int, circuit: cirq.Circuit) -> int:
+    """Add the next non-empty CZ layer and return the updated layer index."""
+    cz_layer = None
+    while not cz_layer:
+        qubits = cast(Iterable[cirq.GridQubit], circuit.all_qubits())
```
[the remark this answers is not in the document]
kenya-sk: I updated `_add_cz_layer()` to take the ordered qubits sequence from the caller instead of reconstructing `circuit.all_qubits()` each time. That makes the data flow clearer, avoids repeated set construction.

kenya-sk:
@mhucka 
Thank you for your review.
I replied your comment and fixed suggestion. Please review this again.

scripts/test_benchmarks.sh
```diff
@@ -14,16 +14,12 @@
 # limitations under the License.
 # ==============================================================================
 echo "Testing all Benchmarks.";
-bazel test -c opt --cxxopt="-msse2" --cxxopt="-msse3" --cxxopt="-msse4" --test_output=errors $(bazel query //benchmarks/scripts:all)
-# test_outputs=$(bazel test -c opt --cxxopt="-msse2" --cxxopt="-msse3" --cxxopt="-msse4" --test_output=errors $(bazel query //benchmarks/scripts:all))
-bench_outputs=$()
-# bench_outputs=$(bazel run -c opt --cxxopt="-msse2" --cxxopt="-msse3" --cxxopt="-msse4" --test_output=errors //benchmarks/scripts:benchmark_clifford_circuit)
+bazel test -c opt --cxxopt="-D_GLIBCXX_USE_CXX11_ABI=1" --cxxopt="-msse2" --cxxopt="-msse3" --cxxopt="-msse4" --test_output=errors $(bazel query //benchmarks/scripts:all)
```
mhucka: `--cxxopt="-D_GLIBCXX_USE_CXX11_ABI=1"` can be removed because it is the default these days.
```suggestion
bazel test -c opt --cxxopt="-msse2" --cxxopt="-msse3" --cxxopt="-msse4" --test_output=errors $(bazel query //benchmarks/scripts:all)
```

mhucka (approved):
Looks good. Thanks for doing this.

I had a very minor change request in one of the scripts, but otherwise this is ready to go.

scripts/test_benchmarks.sh
kenya-sk: Thank you for the thorough review.
I have completed the argument changes to align with the latest updates.

PR #1050, opened 2026-04-17
Refactor Python Toolchain Configuration to use Bazel Repository Rules

ratgr:
## Summary
This PR refactors how Python is detected and configured in TensorFlow Quantum. It moves away from manual script-based file generation for Python paths and adopts a more robust, Bazel-idiomatic approach using repository rules and rules_python toolchains.

## Key Changes 
**Bazel-Managed Python:** Introduced third_party/python_configure.bzl to handle Python detection via a repository rule. This dynamically generates the Python toolchain based on the environment.
- Updated the configuration script to propagate detected paths (PYTHON_BIN_PATH, TF_HEADER_DIR, etc.) to Bazel using -- repo_env. This ensures consistency between the configuration step and the build execution.
- Documentation Updates: Updated docs/install.md to reflect the new configuration flow, including instructions on how to manually specify or override the Python path via --python or --repo_env.

## Fixes multiple python versions:

One of the primary pain points in the previous setup was ambiguity when multiple Python versions (e.g., system Python, pyenv, Conda, and venvs) were present on the same machine. This PR addresses those issues through:

**Strict Path Locking:** Locks the specific interpreter identified during configuration into the Bazel workspace via --repo_env. This prevents "it works in my shell but fails in Bazel" errors caused by path mismatches.

**Explicit Toolchain Registration:** Registers a dedicated py_runtime so Bazel uses the exact same binary for testing, dependency resolution, and execution, removing reliance on ambiguous host environment variables like PYTHONPATH.

**Ambiguity Resolution** via --python: Provides a surgical override through ./configure.sh --python=/path/to/python, allowing developers to bypass auto-detection and explicitly define the interpreter Bazel should use.

**Hermetic Pip Parsing:** Synchronizes the pip_parse interpreter with the build toolchain, ensuring that downloaded wheels and environment scans are binary-compatible with the runtime environment.


third_party/python_configure.bzl
```diff
@@ -0,0 +1,51 @@
...
+        "python-path",
+        "PYTHON_BIN_PATH",
+    ],
+)
+
+def python_configure():
+    """Configures the Python toolchain for TFQ, TF, and XLA."""
+    _python_configure(name = "local_config_python")
+    _python_configure(name = "local_execution_config_python")
+    _python_configure(name = "python")
```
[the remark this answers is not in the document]
ratgr: Added some documentation

docs/install.md
```diff
@@ -184,6 +184,17 @@ build:
 </pre>
 <!-- common_typos_enable -->
 
+The `configure.sh` script detects your Python interpreter and sets up a Bazel toolchain. If you need to manually specify a Python interpreter, you can do so by passing the `--python` flag or by setting the `PYTHON_BIN_PATH` environment variable:
```
mhucka: Thanks for this work!

Minor nit: please keep line lengths to the 80 column convention in this project.

docs/install.md
```diff
@@ -184,6 +184,17 @@ build:
...
+
+<!-- common_typos_disable -->
+<pre class="devsite-click-to-copy">
+  <code class="devsite-terminal">./configure.sh --python=/path/to/python</code>
+</pre>
+<!-- common_typos_enable -->
+
+> [!TIP]
+> You can also bypass manual configuration by passing the Python path directly to Bazel using `--repo_env=PYTHON_BIN_PATH=/path/to/python`.
```
mhucka: Please reformat to 80-columns.

third_party/python_configure.bzl
```diff
@@ -0,0 +1,61 @@
...
+
+"""Repository rule for Python detection and toolchain registration."""
+
+def _python_configure_impl(repository_ctx):
+    python_bin = repository_ctx.os.environ.get("PYTHON_BIN_PATH") or \
+                 repository_ctx.which("python3") or \
+                 repository_ctx.which("python")
+
+    if not python_bin:
+        fail("Python interpreter not found. Please provide it via --repo_env=PYTHON_BIN_PATH=/path/to/python or set the PYTHON_BIN_PATH environment variable.")
```
mhucka: Please reformat to 80-columns.

configure.sh
```diff
@@ -51,37 +51,7 @@ function inside_docker() {
...
-# Absolute path to the python interpreter Bazel/TF should use:
-interpreter = "${PYTHON_BIN_PATH}"
-py_runtime = native.py_runtime
-EOF
-
-  echo
-  echo "Created third_party/python_legacy."
-  echo "Python interpreter = ${PYTHON_BIN_PATH}"
-}
+# Python detection is now handled by the Bazel repository rule in third_party/python_configure.bzl
```
mhucka: Although I understand the reason for the comment, it's better to leave this kind of note to the PR description. Comments that say things like "now handled by" will go out of date – in the future, no one reading this file will remember it was done differently. So, I suggest simply dropping this comment.

configure.sh
```diff
@@ -211,8 +181,7 @@ write_tf_rc "build --repo_env=TF_NEED_CUDA=${TF_NEED_CUDA}"
 # Make sure repo rules and sub-config see legacy Keras (keras 2 instead of Keras 3)
 write_tf_rc "build --repo_env=TF_USE_LEGACY_KERAS=1"
 
-# --- write third_party/python_legacy/ with interpreter --------------------
-write_legacy_python_repo
+# Python detection is now handled by the Bazel repository rule.
```
mhucka: Same comment as above: I suggest dropping this and having this kind of info only live in the PR description.

mhucka (changes requested):
Thanks for this! This is great. I successfully tested it locally. I have some very minor comments. If you could take a look at those, and also update the branch, then I think this will be ready to merge.

mhucka (approved):
Thanks for the updates! Looks good.

PR #1051, opened 2026-04-17
Add support for TensorFlow 2.19

psamanoelton:
## Summary

Update TensorFlow Quantum build to support TensorFlow 2.19.1.

Keep TFQ aligned with TensorFlow 2.19.1’s Bazel and Python repository expectations, confirm Eigen remains on TensorFlow’s pinned revision, and regenerate Python dependencies from `requirements.in`.

### `WORKSPACE`
- bump `@org_tensorflow` to `2.19.1`
- add `python_version_repo` to satisfy TensorFlow 2.19’s expected Python version repository layout
- add lightweight wrapper lockfiles (`requirements_lock_3_10.txt`, `requirements_lock_3_11.txt`, `requirements_lock_3_12.txt`) so the same compiled requirements can be reused across the supported Python versions
- confirm Eigen remains aligned with TensorFlow 2.19.1’s pinned revision

### `requirements.in` / `requirements.txt`
- update TensorFlow requirement to `2.19.1`
- update `tf-keras` requirement to `2.19.x`
- regenerate `requirements.txt` from `requirements.in` with `pip-compile` under Python 3.10
- keep the dependency set compatible with Python 3.10, 3.11, and 3.12

### TFQ source compatibility
- update noisy expectation ops to avoid copy-constructing `tensorflow::mutex`, which fails with TensorFlow 2.19
- keep the rest of the TensorFlow Quantum codepath unchanged as much as possible

### Docs / install flow
- update install docs and tutorial notebook snippets to use TensorFlow `2.19.1`
- document `tf-keras==2.19.0` and `TF_USE_LEGACY_KERAS=1` as part of the supported install path

## Testing

Linux/Ubuntu 22.04 (Docker), Python 3.10, 3.11, and 3.12, Bazelisk → Bazel 6.5.0

- `bazel build ... release:build_pip_package` succeeds
- unit tests (`./scripts/test_all.sh` and `./scripts/ci_validate_tutorials.sh`) pass
- wheel builds, installs, and imports successfully
- validated with TensorFlow `2.19.1` + `tf-keras 2.19.0`, Cirq `1.5.0`, and NumPy `2.1.3`

## Notes

- This keeps TFQ on the Keras 2 path; Keras 3 migration can be handled in a future PR
- TensorFlow 2.19.1 keeps the same Eigen pin already used by TFQ
- this change should help with the incoming upgrades to `2.20`
- Updating Bazel to 7.4.1 might be useful when upgrading to 2.20


scripts/generate_requirements.sh
```diff
@@ -24,17 +24,22 @@ thisdir=$(CDPATH="" cd -- "$(dirname -- "${0}")" && pwd -P)
...
-# lead to pip trying to uninstall an existing version installed by distutils.
-declare -a constraints=()
-constraints+=(--unsafe-package pyyaml)
+if ! python -m pip show -qq pip-tools; then
+  echo "Error: 'pip-compile' not found. Please install 'pip-tools'." >&2
+  exit 1
+fi
 
 # Check for a constraints file and use it if it exists.
+declare -a constraints=()
```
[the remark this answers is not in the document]
psamanoelton: The previous `PyYAML` exclusion does not carry over cleanly with the current `pip-tools` flow we are using here (`--generate-hashes` plus `--allow-unsafe`):

- If I add back `--unsafe-package pyyaml` and keep `--allow-unsafe`, `pyyaml` is still emitted in `requirements.txt` under the unsafe section.
- If I remove `--allow-unsafe` to force `pyyaml` out, `pip-tools` warns that the generated hashed requirements file may be rejected by `pip install`.

Since this PR’s validated target environment is an isolated Docker/venv workflow, and the full build/test cycle passed there, I kept the current behavior for reproducible installs rather than partially restoring the old exclusion.

I agree this is worth revisiting separately if we want to optimize for system-Python/distutils environments again, but I’d prefer to keep this PR scoped to TF 2.19.1 support.

docs/tutorials/mnist.ipynb
```diff
@@ -101,7 +101,7 @@
    "outputs": [],
    "source": [
     "# In Colab, you will be asked to restart the session after this finishes.\n",
-    "!pip install tensorflow==2.18.1 tensorflow-quantum==0.7.6"
+    "!pip install tensorflow==2.19.1 tf-keras==2.19.0 tensorflow-quantum==0.7.6"
```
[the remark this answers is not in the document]
psamanoelton: `TF_USE_LEGACY_KERAS=1` is already set before the first TensorFlow / TensorFlow Quantum import in this notebook, and the same pattern was applied across the tutorials updated in this PR. In `mnist.ipynb`, the install cell is followed by a separate cell that sets:

```python
import os
os.environ["TF_USE_LEGACY_KERAS"] = "1"
```

mhucka:
@psamanoelton Sorry for the delay in getting to this. When you get a chance, could you see about updating the branch, and also resolving the merge conflicts?

scripts/generate_requirements.sh
```diff
@@ -44,10 +49,15 @@ fi
...
-pip-compile -q \
+python -m piptools compile -q \
+  --allow-unsafe \
+  --upgrade \
   --rebuild \
+  --generate-hashes \
   --no-strip-extras \
   --no-emit-index-url \
+  -o requirements.txt \
+  requirements.in \
```
mhucka: pip-compile looks for `requirements.in` automatically, thus it doesn't have to be added as an argument.

docs/tutorials/barren_plateaus.ipynb
mhucka, line 1: The version of `tensorflow-quantum` in the notebooks should be 0.7.7. (That's the next version, and also the working version in GitHub.)

requirements_lock_3_10.txt
mhucka, line 1: I am not sure I understand the purpose of these 3 files, since all they do is load `requirements.txt`.

docs/tutorials/barren_plateaus.ipynb
mhucka, line 1: It should not be necessary to include `tf-keras` here because `tensorflow-quantum` requires tf-keras 2.19. Pip should resolve the dependencies. Is it not doing that for you?

mhucka (commented):
Some preliminary comments. After the current merge conflicts are resolved, I'll take another look.

requirements_lock_3_10.txt
psamanoelton: These three files are intentional.

TensorFlow 2.19's `python_repository(...)` path in `WORKSPACE` expects one lock entry per declared interpreter version:

- `3.10`
- `3.11`
- `3.12`

In our case we want to reuse the same generated `requirements.txt` across those supported versions, so the three `requirements_lock_*` files are just thin wrappers that all point back to `requirements.txt`.

That lets us satisfy the Bazel/TensorFlow repository rule shape without maintaining three separate dependency lockfiles.

docs/tutorials/barren_plateaus.ipynb
psamanoelton: Updated.

I changed the tutorial install cells to:

`pip install tensorflow==2.19.1 tensorflow-quantum==0.7.7`

and removed the explicit `tf-keras` pin.

The extra `tf-keras` install was only needed with the published `0.7.6` wheel, which still carried the older dependency range. In this branch, `0.7.7` already declares `tf-keras>=2.19,<2.20` in `release/setup.py`, so once `0.7.7` is the package being used, pip should resolve it transitively.

mhucka:
@psamanoelton Thank you for the work. 

I'm not sure what's happened with the merges, but right now this PR has a number of things that are not part of the actual changes for the 2.19 support. I have created a new branch and cherry-picked your 2.19-related commits to it, and am currently testing locally. I'll update here when that's done.

mhucka:
Superceded by https://github.com/tensorflow/quantum/pull/1077

Issue #1055, opened 2026-04-18
Use the `README.md` file as the description for `setup()`


mhucka:
In `release/setup.py`, instead of using a hardwired string for the package description in setuptools' `setup()` (which also ends up being the description used on PyPI), we should use the content of the `README.md` file. This is what we do for other projects, and would make the PyPI page more informative.

mhucka:
Resolved by #1057.

PR #1057, opened 2026-04-27
Pypi description load from README

kenya-sk:
Use the project `README.md` as the PyPI long description instead of the hardcoded module docstring in `release/setup.py` (Issue #1055 ). 

Keep image paths in `README.md` relative for repository portability, and rewrite them to absolute GitHub raw URLs only when generating the PyPI long description so they continue to render correctly on PyPI.


kenya-sk:
@mhucka 
I confirmed that the CI was passing. 
I would appreciate it if you could review it.

release/build_pip_package.sh
```diff
@@ -63,6 +63,7 @@ main() {
 
   # Copy over files necessary to run setup.py
   cp "${EXPORT_DIR}/release/setup.py" "${TMPDIR}"
+  cp "${EXPORT_DIR}/release/README.md" "${TMPDIR}"
```
mhucka: If I'm following the code corretly, it looks like the README.md file that will be used for the `long_description` in setup.py is the one from the release/ subdirectory, whereas it should actually use the top-level README.md file. In other words, the text appearing at https://pypi.org/project/tensorflow-quantum/ should be the contents of https://github.com/tensorflow/quantum#readme . 

It may be a simple matter of changing the line above like this:

```suggestion
  cp "${EXPORT_DIR}/README.md" "${TMPDIR}"
```

release/setup.py
```diff
@@ -61,7 +68,25 @@ def read_version():
...
+def read_readme():
+    """Return the project README contents for PyPI."""
+
+    possible_paths = possible_paths_from_setup_dir("README.md")
+    image_src_pattern = re.compile(
+        r'(?P<prefix>src=")(?![a-z]+://|#|/)(?P<path>[^"]+)')
+    image_base_url = ("https://raw.githubusercontent.com/tensorflow/quantum/"
+                      "master/")
+
+    for readme_path in possible_paths:
```
mhucka: To find the README file, the loop could be removed by making use of Python's `next` along the following lines:

```python
possible_paths = possible_paths_from_setup_dir("README.md")
readme_path = next((p for p in possible_paths if p.is_file()), None)
if not readme_path:
        raise RuntimeError("Could not find README.md. Checked:\n" +
                           "\n".join(f"  - {p}" for p in possible_paths))
content = readme_path.read_text(encoding="utf-8")
# ... other code ...
```

release/setup.py
```diff
@@ -61,7 +68,25 @@ def read_version():
...
-DOCLINES = __doc__.split("\n")
+
+def read_readme():
+    """Return the project README contents for PyPI."""
+
+    possible_paths = possible_paths_from_setup_dir("README.md")
+    image_src_pattern = re.compile(
+        r'(?P<prefix>src=")(?![a-z]+://|#|/)(?P<path>[^"]+)')
+    image_base_url = ("https://raw.githubusercontent.com/tensorflow/quantum/"
+                      "master/")
```
mhucka: This kind of approach (regexp to match something in HTML inside a Markdown file) is bound to be fragile and difficult to maintain. I know what you had to do this, though, and I think there is a simpler solution that will avoid the need for this. I will do a PR to replace the relative path to the image in the README.md file to be an absolute path (basically, the path above). This will work for GitHub and PyPI and save the need for the regex matching and substitution.

mhucka (changes requested):
Thank you for working on this!

I think there may be a small error in which README.md file gets used. Also, I will do a quick PR to replace the relative path to the logo image in the top-level README.md file so that it's an absolute path, and this will avoid the need for the regex-based replacement implemented in this PR.

release/setup.py
mhucka: Update: the PR to change the URLs is https://github.com/tensorflow/quantum/pull/1069

release/build_pip_package.sh
kenya-sk: Thank you for pointing that out.
I misunderstood. I've fix it.

release/setup.py
kenya-sk: The README is now loaded from the project root. As a result, the loop processing was no longer necessary and has been removed.

mhucka (approved):
LGTM

PR #1070, opened 2026-05-13
Fix C++ int type mismatch warnings in core/ops files

mhucka:
PR https://github.com/tensorflow/quantum/pull/476 (dating from 2021) aimed to resolve numerous warnings emitted by the C++ compiler. Most have already been resolved since the time the PR was written, but some still remained. This PR addresses most (possibly all) the rest that can be addressed. (Some cannot be addressed directly because they come from other libraries.)

All of these concern the core ops (`tensorflow_quantum/core/ops/`) and all the warnings are essentially in the same category of thing, so while this PR touches many files, it seems appropriate to handle them together:

* Changed the data type of variables `max_num_qubits` and `min_num_qubits` from `int` to `uint64_t` to match the underlying data structures and vector iterations where they were being used.

* Changed the data types of some iterator and limit variables (`int q_ind`, `int64_t q_ind`, `int nq`, and `int largest_nq`) to `uint64_t` to resolve compiler warnings when evaluated against the updated `max_num_qubits` mentioned above.

* Changed the data type of `output_dim_size` from `int` to `size_t` where this variable was directly assigned the result of `maps.size()` and similar standard library size methods.

* Retained the data types of variables like `max_psum_length` have been left as `int` because they interface with Protobuf sizes, which are signed ints.

* In `configure.sh`, removed `-Wno-unused-but-set-variable` from the flags used during builds of `tensorflow_quantum/core/ops/noise/tfq_.*`, since the underlying reason for the warnings has been addressed by this PR.


MichaelBroughton (approved):
LGTM

Issue #908, opened 2025-09-16
Add pre-commit hook support for code quality and consistency


Aniketsy:
The repository currently does not have pre-commit hook support. Adding pre-commit would help automate code quality checks, enforce consistent formatting, and catch common issues before code is committed. Implementing pre-commit hooks could improve code reliability and make collaboration easier for all contributors.

I would like to suggest adding this, but if the maintainers feel it is not needed, please let me know.

PR #1071, opened 2026-05-13
Fix miscellaneous trivial typos

mhucka:
This PR fixes random typos in the text of about three dozen files. The typos were flagged by `codespell`, which I'm beginning to use in my own pre-commit workflows. This is admittedly trivial stuff, but we may as well fix these silly typos in the source and be done with it. With these out of the way, we'll also be able to address https://github.com/tensorflow/quantum/issues/908 (request for a pre-commit hooks configuration) with a codespell check as one of the hooks.


MichaelBroughton (approved):
LGTM

PR #1072, opened 2026-05-18
Merged version of TF 2.19 PR

mhucka:


mhucka:
Closing so that Pablo can open a new one

PR #1074, opened 2026-05-22
Use r-strings for docstrings containing LaTeX markup

mhucka:
A few docstrings in Python files contained LaTeX mathematical notation. They should be made into raw strings (using the `r` prefix) so that Python doesn't misinterpret the backslash characters.

There are no code or content changes in this PR.


MichaelBroughton (approved):
Might want to double check how these render on the site.

Issue #1066, opened 2026-05-12
Test that channel-based random circuits are correctly generated and handled


mhucka:
PR https://github.com/tensorflow/quantum/pull/1056 currently tests only verify the default behavior where `include_channels=False`. It would be beneficial to add test cases or parameterize these tests to also cover `include_channels=True` to ensure that channel-based random circuits are correctly generated and handled.

rosspeili:
Hey @mhucka can I give this a shot?

mhucka:
@rosspeili Circling back to this: if you're still interested, we'll assign it to you.

rosspeili:
Yes @mhucka

rosspeili:
@mhucka Opened #1091 for this, waiting for ci checks to pass, and see if it needs adjustments. Followed the Gemini assistant's suggestions, let me know how it looks and if there's anything I am missing. Thanks.

PR #1080, opened 2026-06-03
Add channel random circuit test coverage

arnavnagzirkar:
Adds regression coverage for random circuit generation with channels.

The `include_channels=True` argument mentioned here is a parameter of two functions in the `tensorflow_quantum.python.util` module (imported as `util` in the tests):

* `util.random_circuit_resolver_batch`
* `util.random_symbol_circuit_resolver_batch`

The new tests in `tensorflow_quantum/python/util_test.py` assert that, when each of these functions is called with `include_channels=True`, the generated random circuits contain at least one supported channel operation (taken from `util.get_supported_channels()`).

Following the Gemini Code Assist review, the tests also verify that the circuits survive a full serialization round trip: `util.convert_to_tensor` is used to serialize, `util.from_tensor` to deserialize, and the re-serialized result is compared for equality. This is stronger than only checking that serialization does not return null.

This addresses issue #1066.

### DIFF
--- a/tensorflow_quantum/python/util_test.py
+++ b/tensorflow_quantum/python/util_test.py
@@ -159,8 +159,15 @@ class UtilFunctionsTest(tf.test.TestCase, parameterized.TestCase):
             any(has_channel(c) for c in circuits),
             "Expected at least one channel operation across batch when "
             "include_channels=True")
-        # Verify circuits containing channels are serializable.
-        self.assertIsNotNone(util.convert_to_tensor(circuits))
+        # Verify circuits containing channels are correctly serializable
+        # round-trip.
+        serialized = util.convert_to_tensor(circuits,
+                                            deterministic_proto_serialize=True)
+        deserialized = util.from_tensor(serialized)
+        self.assertAllEqual(
+            serialized,
+            util.convert_to_tensor(deserialized,
+                                   deterministic_proto_serialize=True))
 
     def test_random_symbol_circuit_resolver_batch_channels_present(self):
         """Confirm channel ops appear in circuits when include_channels=True."""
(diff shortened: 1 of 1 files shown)


mhucka:
@arnavnagzirkar Thank you for this contribution. Gemini Code Assist's review comments seem reasonable. If you don't disagree with them, could you got ahead and make those changes?

Also, in the PR description, it talks about include_channels=True, but it doesn't mention which function or module that refers to. Could you add that information to the description?

arnavnagzirkar:
Thanks @mhucka. Two quick updates:

1. The Gemini round trip suggestion is already in the branch. I applied it in commit 9b1f9a8, which I pushed before your comment, so it was probably just not obvious that the threads were already handled. Both new tests now serialize with `util.convert_to_tensor(..., deterministic_proto_serialize=True)`, deserialize with `util.from_tensor`, and assert that re-serializing the deserialized circuits gives back the same tensor. That makes it a full round trip rather than the earlier not-null check.

2. I updated the PR description to say where `include_channels=True` comes from. It is a parameter of `util.random_circuit_resolver_batch` and `util.random_symbol_circuit_resolver_batch` in the `tensorflow_quantum.python.util` module.

Let me know if you would like anything else adjusted.

tensorflow_quantum/python/util_test.py
```diff
@@ -139,6 +139,65 @@ def test_random_symbol_circuit_resolver_batch_shapes_and_types(
                     isinstance(value, float)
                     for value in resolver.param_dict.values()))
 
+    def test_random_circuit_resolver_batch_channels_present(self):
+        """Confirm channel ops appear in circuits when include_channels=True."""
+        qubits = cirq.GridQubit.rect(1, 3)
+        channel_types = tuple(type(c) for c in util.get_supported_channels())
+
+        circuits, _ = util.random_circuit_resolver_batch(qubits,
+                                                         batch_size=5,
```
mhucka, line 148: For this and the next new function, could you please use a variable to set the batch size, following the pattern of the other functions in this file? (See for example this line: https://github.com/tensorflow/quantum/pull/1080/changes#diff-0605f6d13dea49e94d3ddf78736d228215528e5f046a281dfddc1ad9afd63cddR119)

mhucka (changes requested):
Thank you for working on this. I have a small change request, but otherwise this looks good and almost ready to merge.

arnavnagzirkar:
Thanks @mhucka for the review and for the time on this.

I see that #1091 was merged in the meantime and already adds comprehensive channel coverage for the random circuit generators, closing issue #1066. Its tests (test_random_circuit_resolver_batch_channel_content, test_random_symbol_circuit_resolver_batch_channel_content, and test_random_symbol_circuit_channel_content) cover everything in this PR and more, including the include_channels=False path and validation that channel ops are not controlled or parameterized.

Since the two tests here are now a subset of what is already on master, I am closing this PR to avoid duplicate coverage. Happy to help elsewhere if there is a follow up you would like. Thanks again.

mhucka:
@arnavnagzirkar Whoops, I didn't even notice that the PRs overlapped so much. Sorry about that!  (This is a hazard of reviewing PRs for a half dozen different projects …)

Thank you for your offer to work on other things. I recently [tagged more issues with "help wanted"](https://github.com/tensorflow/quantum/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22help%20wanted%22) – perhaps there is something there that would spark your interest?

PR #1091, opened 2026-06-06
Test channel inclusion in random circuit generators.

### LINKED ISSUE #1066: Test that channel-based random circuits are correctly generated and handled

rosspeili:
Behavioral coverage for include_channels=True/False in util_test, including channel presence, non-parameterization, and TFQ serialization round-trips. Document include_channels in random circuit helper docstrings.

Fixes #1066

### DIFF
--- a/tensorflow_quantum/python/util_test.py
+++ b/tensorflow_quantum/python/util_test.py
@@ -53,20 +53,44 @@ def _supported_channel_types():
     return tuple(type(channel) for channel in util.get_supported_channels())
 
 
-def _is_supported_channel(gate):
+def _op_base_gate_and_controlled(op):
+    """Return the underlying gate and whether the op is controlled."""
+    is_controlled = False
+    sub_op = op
+    if isinstance(op, cirq.ControlledOperation):
+        sub_op = op.sub_operation
+        is_controlled = True
+
+    gate = sub_op.gate
+    if isinstance(gate, cirq.ControlledGate):
+        gate = gate.sub_gate
+        is_controlled = True
+
+    return gate, is_controlled
+
+
+def _op_has_supported_channel(op):
+    gate, _ = _op_base_gate_and_controlled(op)
     return isinstance(gate, _supported_channel_types())
 
 
 def _circuit_has_supported_channel(circuit):
     return any(
-        _is_supported_channel(op.gate) for moment in circuit for op in moment)
+        _op_has_supported_channel(op) for moment in circuit for op in moment)
 
 
-def _assert_channels_not_parameterized(testcase, circuit):
+def _assert_channel_ops_valid(testcase, circuit):
+    """Assert channels are neither parameterized nor controlled."""
     for moment in circuit:
         for op in moment:
-            if _is_supported_channel(op.gate):
-                testcase.assertFalse(cirq.is_parameterized(op))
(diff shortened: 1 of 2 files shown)


tensorflow_quantum/python/util_test.py
```diff
@@ -47,6 +49,26 @@ def _exponential(theta, op):
...
+def _circuit_has_supported_channel(circuit):
+    return any(
+        _is_supported_channel(op.gate) for moment in circuit for op in moment)
+
+
+def _assert_channels_not_parameterized(testcase, circuit):
+    for moment in circuit:
+        for op in moment:
+            if _is_supported_channel(op.gate):
+                testcase.assertFalse(cirq.is_parameterized(op))
```
mhucka: I think there's a subtle flaw there:

1. `random_symbol_circuit` in util.py can call `_apply_random_control`
2. `_apply_random_control` may wrap a gate inside `cirq.ControlledOperation`
3. the test above checks if op.gate is a supported channel type. However, op.gate can be a `ControlledGate` wrapper …
4. which in `_is_supported_channel` will be tested against raw channel classes …
5. which will return False …
6. which means the test `cirq.is_parametrized` inside the `if` condition will not be tested …
7. which means those cases are silently ignored by the test. 

So I think the function needs to be changed to something like the following:

```python
def _assert_channels_not_parameterized_or_controlled(testcase, circuit):
    """Asserts that channels are neither parameterized nor controlled."""
    for moment in circuit:
        for op in moment:
            sub_op = op
            is_controlled = False

            if isinstance(op, cirq.ControlledOperation):
                sub_op = op.sub_operation
                is_controlled = True

            gate = sub_op.gate
            if isinstance(gate, cirq.ControlledGate):
                gate = gate.sub_gate
                is_controlled = True

            if isinstance(gate, _supported_channel_types()):
                testcase.assertFalse(
                    is_controlled,
                    msg=f"Found controlled channel operation: {op}"
                )
                testcase.assertFalse(
                    cirq.is_parameterized(op),
                    msg=f"Found parameterized channel operation: {op}"
                )
```

The calls to `_assert_channels_not_parameterized` also should be changed to the new function. (Maybe a shorter name can be found, because the one above is kind of ridiculously long, but right now I can't think of a better one.)

mhucka (changes requested):
Thank you for this work!

Please see the comment about the test that may miss some possible cases.

Also, I did not have time to look further down to see if there's a similar situation elsewhere in util_test.py, so please check that.

tensorflow_quantum/python/util_test.py
rosspeili: Thanks for the direction @mhucka, and you were right to assume similar gaps. Same unwrapping logic applied also to the channel presence helper. Channel ops are now checked as neither controlled nor parameterized via `_assert_channel_ops_valid`. Let me know how this looks, or if you want a better name for the function. <3

mhucka (approved):
Thanks for the changes. Looks good to me now!

PR #1097, opened 2026-06-22
Fix vulnerability: uninitalized tensor heap memory leak via integer overflow in BalanceTrajectory

mhucka:
Internal security scans reported a security vulnerability (CWE-200/CWE-908) in `tensorflow_quantum/core/ops/noise/tfq_noisy_samples.cc`. In `TfqNoisySamplesOp`, a massive `num_samples` value near `INT_MAX` will cause `BalanceTrajectory` to overflow its signed integer calculation for chunk sizes. This results in negative or zero chunk sizes, causing worker threads to bypass execution. Because TensorFlow's `allocate_output` provides uninitialized heap blocks, the kernel returns a tensor filled with residual process memory. 

Changes in this PR:

* In `util_balance_trajectory.cc`, updated chunk calculations to use `int64_t` arithmetic, preventing overflows when `num_samples` (or `rep_limits`) is close to `INT_MAX`.

* Modified `tfq_noisy_samples.cc`, `tfq_noisy_expectation.cc`, and         `tfq_noisy_sampled_expectation.cc` to compute thread allocations (`p_reps`) using `int64_t` division and cast the  Philox random sample reservation requests to `int64_t` variables.

* Add checks to enforce a maximum capacity limit of 10,000,000 in input tensor parsing functions `GetNumSamples` and `GetIndividualSample` within `parse_context.cc`. This should eliminate memory depletion/DoS attempts with massive sizes.

* Added tests.

### DIFF
--- a/tensorflow_quantum/core/ops/noise/tfq_noisy_expectation.cc
+++ b/tensorflow_quantum/core/ops/noise/tfq_noisy_expectation.cc
@@ -119,8 +119,8 @@ class TfqNoisyExpectationOp : public tensorflow::OpKernel {
 
     Status parse_status = ::tensorflow::Status();
     auto p_lock = absl::Mutex();
-    auto construct_f = [&](int64_t start, int64_t end) {
-      for (int64_t i = start; i < end; i++) {
+    auto construct_f = [&](int start, int end) {
+      for (int i = start; i < end; i++) {
         Status local = NoisyQsimCircuitFromProgram(
             programs[i], maps[i], num_qubits[i], false, &qsim_circuits[i]);
         NESTED_FN_STATUS_SYNC(parse_status, local, p_lock);
@@ -132,8 +132,8 @@ class TfqNoisyExpectationOp : public tensorflow::OpKernel {
         programs.size(), num_cycles, construct_f);
     OP_REQUIRES_OK(context, parse_status);
 
-    uint64_t max_num_qubits = 0;
-    for (const uint64_t num : num_qubits) {
+    int max_num_qubits = 0;
+    for (const int num : num_qubits) {
       max_num_qubits = std::max(max_num_qubits, num);
     }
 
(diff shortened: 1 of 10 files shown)


tensorflow_quantum/core/ops/noise/tfq_noisy_expectation.cc
```diff
@@ -132,8 +132,8 @@ class TfqNoisyExpectationOp : public tensorflow::OpKernel {
         programs.size(), num_cycles, construct_f);
     OP_REQUIRES_OK(context, parse_status);
 
-    uint64_t max_num_qubits = 0;
-    for (const uint64_t num : num_qubits) {
+    int max_num_qubits = 0;
```
MichaelBroughton, line 135: Can we avoid downgrading these ?

tensorflow_quantum/core/ops/noise/tfq_noisy_samples.cc
```diff
@@ -60,7 +61,10 @@ class TfqNoisySamplesOp : public tensorflow::OpKernel {
 
   void Compute(tensorflow::OpKernelContext* context) override {
     // TODO (mbbrough): add more dimension checks for other inputs here.
-    DCHECK_EQ(4, context->num_inputs());
+    const int num_inputs = context->num_inputs();
+    OP_REQUIRES(context, num_inputs == 4,
+                tensorflow::errors::InvalidArgument(absl::StrCat(
```
MichaelBroughton, line 66: This seems unrelated to fixing the overflow problem, can we move this to a different PR ?

tensorflow_quantum/core/ops/noise/tfq_noisy_samples.cc
```diff
@@ -79,15 +83,19 @@ class TfqNoisySamplesOp : public tensorflow::OpKernel {
 
     int num_samples = 0;
     OP_REQUIRES_OK(context, GetIndividualSample(context, &num_samples));
+    OP_REQUIRES(context, num_samples >= 0 && num_samples <= 10000000,
+                tensorflow::errors::InvalidArgument(
```
MichaelBroughton, line 87: This seems like the main piece of code that actually solves our problems, can we scope this PR down to just putting in these checks and testing them across all ops that require a num_samples and call into BalanceTrajectory.

MichaelBroughton (changes requested):
We can probably shorten this up a fair bit and remove some unrelated changes that didn't get new tests. Fine to merge once this is done.

PR #1100, opened 2026-06-30
Unpin clang-format in development requirements.

### LINKED ISSUE #1084: Unpin clang-format version
At some point in the recent past, we pinned the version of clang-format to version 18 in the requirements files. However, it looks like this is unnecessary; running a more recent version of clang-format like 22 does not produce any changes to formatting. We should unpin the version in `requirements.in`  if we can.

rosspeili:
Remove the ~=18.1.0 constraint from requirements.in and refresh the locked clang-format entry in requirements.txt to 22.1.5. Closes #1084


mhucka (approved):
Well, that was easier than expected. Thanks!

